PR #83, opened 2014-12-27
Adding a widget for displaying MOC information

anticlockwise:
This is a widget for displaying MOC (Music On Console) information just like the MPD widget. Also provides naughty notifications on song change.


copycat-killer:
Why don't you try to implement it like described in #86?

Thinking better, a clone is not an elegant solution.

PR #86, opened 2015-01-02
Adding cmus widget

aajjbb:
Pulling a widget dedicated to cmus, it currently has options to display the state of the player and current song information.

### DIFF
--- a/widgets/cmus.lua
+++ b/widgets/cmus.lua
@@ -1,0 +1,85 @@
+--[[
+                                                   
+     Licensed under GNU General Public License v2  
+      * (c) 2015, aajjbb <http://github.com/aajjbb>
+                                                   
+--]]
+
+local helpers = require("lain.helpers")
+local async   = require("lain.asyncshell")
+
+local escape_f = require("awful.util").escape
+local naughty  = require("naughty")
+local wibox    = require("wibox")
+
+local io     = { popen   = io.popen }
+local os     = { execute = os.execute,
+                 getenv  = os.getenv }
+local string = { format  = string.format,
+                 gmatch  = string.gmatch }
+
+local setmetatable = setmetatable
+
+local cmus = {}
+
+local function worker(args)
+    local args        = args or {}
+    local timeout     = args.timeout or 2
+    local music_dir   = args.music_dir or os.getenv("HOME") .. "/Music"
+    local cover_size  = args.cover_size or 100
+    local default_art = args.default_art or ""
+    local settings    = args.settings or function() end
+
+    cmus.widget = wibox.widget.textbox('')
+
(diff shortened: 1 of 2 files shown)


copycat-killer:
I suggest to implement it using [abase](https://github.com/copycat-killer/lain/wiki/abase).

Just set the cmus command as `cmd` and adapt the body of your function in `settings`.

You can then pull request me a wiki entry (in the users contributed section) with this use case as a reference. Alternatively, you can post it here and I'll add it.

Let's have a saner trend than just cloning existing widgets behaviour.

aajjbb:
Hi,

Using the abase widget, my parsing logic would be inside my rc.lua, isn't it ?

aajjbb:
Hi,

Using abase to implement, I would have to implement my parsing logic inside rc.lua, isn't it ? It would pollute our rc file, or you mean't I should implement this abase widget inside my cmus widget ?

I've just used other music widget as a base to use with my music player, I didn't meant to do something bad.

copycat-killer:
I didn't say it's bad.

I just wanted to make you exploit abase to achieve that elegantly, instead of adding redundancy.

And yes, I just meant to put the 'logic' inside settings function.

aajjbb:
Great, I can do this, but what do you think about merging the music player widgets so that we pass the player's name as an argument ? Do you think an music player widget unnecessary ?

copycat-killer:
You can use exclusively `abase` or `base` for this. No need for a new widget.

copycat-killer:
There you go.

```lua
cmuswidget = lain.widgets.abase({
    cmd = "cmus-remote -Q",
    settings = function()
        cmus_now = {
            state   = "N/A",
            artist  = "N/A",
            title   = "N/A",
            album   = "N/A"
        }

        for w in string.gmatch(output, "(.-)tag") do
            a, b = w:match("(%w+) (.-)\n")
            cmus_now[a] = b
        end

        -- customize here
        widget:set_text(cmus_now.artist .. " - " .. cmus_now.title)
    end
})
```

copycat-killer:
Added to [abase wiki entry](https://github.com/copycat-killer/lain/wiki/abase) as reference.

kajzersoze:
Seems like abase is not longer available, I'm interested in cmus widget, with what can I replace the above code (instead of abase) ?

lcpz:
@kajzersoze `abase` has been replaced by `awful.widget.watch` a long time ago. All of its recipes, including cmus one, have been put [here](https://awesomewm.org/recipes/watch).

kajzersoze:
I know it's old but I just started using AwesomeWM.
cmus recipe from the link you provided works great, many thanks!

PR #88, opened 2015-01-10
Uselesstile and uselessfair fixes

worron:
Found that useless gap layouts work wrong with thick window borders. While fixing it I decide rewrite tiling code a bit. Please check if this changes suitable for lain module.


copycat-killer:
Yes, it's a good work.

PR #91, opened 2015-02-03
Fix in calendar

nsarfati:
Hi! I'm doing this change because in some distributions (i.e ubuntu 14.04) without the regex clean you can't appreciate the highlight of the current date and you see strange symbols.

Look attached screenshots:

Previous to the change:
previous

After the change:
after


copycat-killer:
I pushed another commit also to remove useless lines.

nsarfati:
But that another commit fix the problem with the date? Look, I just cleaning the output of cal with the same regex that you execute 3 lines down.

copycat-killer:
Look, just read the commit and you'll see. :-)

PR #97, opened 2015-02-16
Add a card attribute for alsa and alsabar

mvalkon:
Currently alsa.lua and alsabar.lua do not support the amixer command
line option "-c" which allows defining which card to use. For at
least new Dell E-series laptops, the card is not the default (0) and
therefore this functionality is required for these widgets to work.


copycat-killer:
Thanks.

PR #106, opened 2015-04-25
fix(async): don't use filesystem

actionless:
I tested it only with the weather widget (`yawn`) but it should work for the rest of them as well.

### DIFF
--- a/asyncshell.lua
+++ b/asyncshell.lua
@@ -3,6 +3,7 @@
                                                   
      Licensed under GNU General Public License v2 
       * (c) 2013, Alexander Yakushev              
+      * (c) 2014, Yauhen Kirylau
                                                   
 --]]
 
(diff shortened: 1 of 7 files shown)


copycat-killer:
`async`uses `/tmp` folder, which is flushed periodically or at boot.

What's the benefit of your proposal?

Can you provide a `strace` and/or `time` diff?

actionless:
`/tmp` is not cleaned by default in some distributions

also not all the users how mounted `/tmp` to ramfs 

i will also try to do some benchmarking

copycat-killer:
`/tmp` is cleaned at boot or by age by every distribution, and it's not necessary to mount it on ramfs.

I'll wait these benchmarks.

actionless:
on boot -- yes, but not in every distro that happens on timer, but whatever.

so the benchmark:
```lua
local socket = require("socket")
local awesome_context = {}
local benchmark_max = 600
local function benchmark_function(async, id)
  awesome_context[id] = 0;
  local start_time = socket.gettime()
  local start_cpu_time = os.clock()
  for i = 1,benchmark_max do
    async.request('echo 1', function(str)
      if id=='b1' then
       str:read("*a")
      end
      awesome_context[id] = awesome_context[id] + 1
      if awesome_context[id] >= benchmark_max then
        local result_time = socket.gettime() - start_time
        local result_cpu_time = os.clock() - start_cpu_time
        print(id .. ":")
        print(awesome_context[id])
        print(result_time)
        print(result_cpu_time)
        awesome_context[id] = 0
      end
    end)
  end
end
```

in average cpu time about 50 ms (15%) less, but realtime about 100 ms (5%) more

so it can be good for responsibility, but i am not sure about how the magic with escaping result output will work in cases with some tricky output so that in particular can require more testing

PR #111, opened 2015-06-20
Optionally disable mapped volume output

Unode:
The ``-M`` option in amixer may be unintuitive for some users.
Allow disabling it by passing an empty string as the value of mapped.

The default behaviour is to use ``-M`` to keep backwards compatibility.

### DIFF
--- a/widgets/alsa.lua
+++ b/widgets/alsa.lua
@@ -28,11 +28,12 @@ local function worker(args)
 
     alsa.card    = args.card or "0"
     alsa.channel = args.channel or "Master"
+    alsa.mapped  = args.mapped or "-M"
 
     alsa.widget = wibox.widget.textbox('')
 
     function alsa.update()
-        local f = assert(io.popen(string.format("amixer -c %s -M get %s", alsa.card, alsa.channel)))
+        local f = assert(io.popen(string.format("amixer -c %s %s get %s", alsa.card, alsa.mapped, alsa.channel)))
         local mixer = f:read("*a")
         f:close()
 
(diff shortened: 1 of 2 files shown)


copycat-killer:
The -M option in amixer may be unintuitive for some users.

Why?

And how it justifies this pull?

Unode:
The ``-M`` option is supposed to make the volume level behave in a more "natural way for the human hear". The same is true for alsamixer. However, I find it unintuitive to have an exponential volume scale and harder to predict the expected volume level given a percentage and found myself creating this patch.

Since the pull doesn't affect existing users and merely allows disabling said behaviour when required, it shouldn't pose as a major problem. No?

copycat-killer:
I know what `-M` does and I saw that your patch is backward compatible.

I just wanted to understand why you proposed it, since in *my* experience that argument has never been a [remarkable difference](https://i.imgur.com/Rvlt4gH.png). 

Moreover, your pull request is incomplete: you should patch `alsabar` too, as well as update the wiki.

If you prove that a difference exists for your system, I'll introduce a more general modification.

Unode:
Hi Luke,

You screenshot is interesting. In my system the volume level "curve" is [completely different](http://www.pasteall.org/pic/show.php?id=89710) when using ``-M``.

Unode:
Updated pull request to include ``alsabar``.

I'm not sure how should I go about editing the wiki as it lives as a submodule.

copycat-killer:
If you prove that a difference exists in your system, I'll introduce a more general modification.

Now amixer call can be customized with additional arguments. Check [wiki](https://github.com/copycat-killer/lain/wiki/alsa) for the details.

I decided to remove `-M` since it's not as default as it seemed to me.

Unode:
Thanks Luke

PR #117, opened 2015-07-11
fixing small typo in fs widget which lead to nil reference

aajjbb:
In the last commit, there was a typo in lines 109/110 from ```fs``` widget which trying to add events to a 'null' widget. It happens because actual widget belong to ```fs``` table, and these lines didn't had the ```fs.``` prefix to the widget. 

Pulling small fix to it.


copycat-killer:
Thanks.

PR #118, opened 2015-07-12
Respect card number setting when fetching volume and invoking alsamixer.

k3rni:


copycat-killer:
Use a custom `command` variable [instead](https://github.com/copycat-killer/lain/wiki/alsabar).

I removed `card` because it gave issues for those who just have one card (which is the majority).

k3rni:
Fair enough, using a custom command solves the problem. Although my case was not multiple cards, but one that shows as two devices: one for speaker output, another for HDMI audio-out (on a laptop), in the wrong order. I suspect the majority might find themselves there too.

copycat-killer:
> I suspect the majority might find themselves there too.

I meant to say that a regular user normally have just one audio output. At least, that's what I see the most.

PR #122, opened 2015-07-23
Asyncshell now works without temporary files

### LINKED ISSUE #120: asyncshell (via mpd widget) leaves (a lot of) files behind in /tmp/asyncshell
I noticed this after enabling the mpd widget that uses asyncshell.

In a machine that is on almost 24/7 there were over 100k files in `/tmp/asyncshell` totalling ~40Mb. `/tmp/` is configured to use tmpfs.
One file is created every time the mpd widget updates which is about once per second.

These files are named `reqXXXXX` where `XXXXX` seems to be some kind of incremental identifier.

worron:
Asyncshell now works without saving data on disk.
Demand function removed.
Optional argument 'timeout' added to prevent outdated results.
Fixes #120

### DIFF
--- a/asyncshell.lua
+++ b/asyncshell.lua
@@ -43,10 +43,9 @@ function asyncshell.request(command, callback, timeout)
     asyncshell.request_table[id] = { callback = callback }
 
     local formatted_command = string.gsub(command, '"','\"')
-                              .. " | sed -e 's/\"/\\\\\"/g' -e ':a;N;$!ba;s/\\n/\\\\n/g'"
 
     local req = string.format(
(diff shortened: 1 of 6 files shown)


copycat-killer:
`imap` and `mpd` doesn't work for me with this version.

worron:
Sorry,  did not make proper test for complex command. It seems fixed now.

copycat-killer:
Due to more recent commits, there were conflicts. Plus, you didn't update `weather` which uses `asynchell` too.

Merged and solved in 1c05dff.

Issue #125, opened 2015-07-27
Widget positioning with multiple monitors


Unode:
Hi there,

When using a multi monitor/screen setup (say 2), if the widget in the second screen is hovered, the "popup" shows up on the first.
Seen with, for instance, `widgets.fs` and `widgets.calendar`.

Is there any way to customize this?

copycat-killer:
Yes, you can customize screen number popup for every supported widget.

For `fs`, set `fs_notification_preset.screen = yournumber` within the `settings` function.

For, `calendar`, set `scr_pos` as input argument.

Read the wiki for the rest.

Unode:
Thanks for the references.

Would it be possible to have these arguments dynamically set (or modified) based on the mouse or the widget location (screen)?

Wondering how to get the popups on the same screen as the widget without having to create new/custom sets of widgets for every screen.

Thanks.

copycat-killer:
It should be possible like this:

```lua
local mouse = mouse

-- [...]

notification_preset.screen = mouse.screen
```

Unode:
Yup that works.

Pull request linked.

PR #126, opened 2015-07-27
Add behaviour to show notification windows on same screen as widget

Unode:
Feature discussed in #125

### DIFF
--- a/widgets/calendar.lua
+++ b/widgets/calendar.lua
@@ -15,6 +15,7 @@ local naughty      = require("naughty")
 local io           = { popen = io.popen }
 local os           = { date = os.date }
 local tonumber     = tonumber
+local mouse        = mouse
 
 local setmetatable = setmetatable
 
@@ -89,6 +90,10 @@ function calendar:show(t_out, inc_offset, scr)
              .. "</span></tt>"
     f:close()
 
+    if calendar.followmouse == true then
+        scr = mouse.screen
+    end
+
     cal_notification = naughty.notify({
         text = c_text,
         icon = calendar.notify_icon,
@@ -112,6 +117,7 @@ function calendar:attach(widget, args)
     calendar.bg        = args.bg or beautiful.bg_normal or "#FFFFFF"
     calendar.position  = args.position or "top_right"
     calendar.scr_pos   = args.scr_pos or 1
+    calendar.followmouse = args.followmouse or false
 
     calendar.offset = 0
     calendar.notify_icon = nil
--- a/widgets/calendar.lua
+++ b/widgets/calendar.lua
@@ -31,9 +31,10 @@ function calendar:hide()
     end
 end
 
-function calendar:show(t_out, inc_offset, scr)
+function calendar:show(t_out, inc_offset)
     calendar:hide()
 
+    local scr  = calendar.scr_pos
     local offs = inc_offset or 0
     local tims = t_out or 0
     local f, c_text
@@ -101,7 +102,7 @@ function calendar:show(t_out, inc_offset, scr)
         fg = calendar.fg,
         bg = calendar.bg,
         timeout = tims,
-        screen = scr or 1
+        screen = scr
     })
 end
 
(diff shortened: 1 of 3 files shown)


Unode:
The pull request also includes a fix for the usage of the `src_pos` argument.
It was incorrectly passed and always ended up as `nil` and then set to default `1`.

Unode:
Wiki updated as per pull-request.

aajjbb:
Great job.

copycat-killer:
I don't understand the need for this.

If you want the `followmouse` behaviour, just set `screen/scr_pos = mouse.screen`.

And about `scr_pos`: `calendar:show` has [`scr` as argument](https://github.com/copycat-killer/lain/blob/master/widgets/calendar.lua#L33), which is [correctly passed](https://github.com/copycat-killer/lain/blob/master/widgets/calendar.lua#L119) for mouse hovering signal, and which can be optionally defined by user.

copycat-killer:
The only thing I'm agree with is updating `show` in supported widgets, in order to define which `screen` to be show in.

Unode:
Where would I set `screen = mouse.screen`? If it's in the widget initialization code, I tired it and it didn't work. It's the only part I can control from `awesome/rc.lua`.
The reference to `mouse.screen` must exist inside the `show` functions of the widget. `mouse.screen` is an integer, doing `screen = mouse.screen` will store whatever value was present at the time of evaluation. It won't update in the future, i.e. not a reference, so it won't actually follow the mouse.

The extra parameter was to maintain backwards compatibility. Not my call to break it.

As for `scr` being correctly passed, the scope is wrong. [`calendar.scr_pos`](https://github.com/copycat-killer/lain/blob/master/widgets/calendar.lua#L114)  holds the value, not [`src_pos`](https://github.com/copycat-killer/lain/blob/master/widgets/calendar.lua#L119). As it is, the calls to [`show`](https://github.com/copycat-killer/lain/blob/master/widgets/calendar.lua#L119) only pass `nil`. Confirmed by printing `scr` inside the function.
Also, you are setting the default value of `src_pos` twice. Once in [`calendar.worker`](https://github.com/copycat-killer/lain/blob/master/widgets/calendar.lua#L114) and again in [`calendar.show`](https://github.com/copycat-killer/lain/blob/master/widgets/calendar.lua#L99) (the `or 1` part). If you drop [this one](https://github.com/copycat-killer/lain/blob/master/widgets/calendar.lua#L99) you'll see the `nil`.

copycat-killer:
Ok, I understood now. But since you didn't update all the required widgets and there are now conflicts with this branch, I'll do the required changes.

By the way,

> Where would I set `screen = mouse.screen`? If it's in the widget initialization code, I tired it and it didn't work. It's the only part I can control from `awesome/rc.lua`.

Minimal working examples:

```lua
lain.widgets.calendar.attach(mytextclock, {scr_pos = mouse.screen})

myfs = lain.widgets.fs({
    notification_preset.screen = mouse.screen
})

myweather = lain.widgets.weather({
    city_id  = 1234567, --placeholder
    notification_preset = { screen = mouse.screen }
})
```

copycat-killer:
Done. Please take a look at e5e199c and see if everything is ok.

Notice that now `fs` has `notification_preset` as entry (no more global), and `args.w_notification_preset` is now `args.notification_preset` in `weather`.

Unode:
Hi Luke,

The examples you mentioned didn't work for me:
```
myfs = lain.widgets.fs({
    notification_preset.screen = mouse.screen
})
```
gives me a syntax error.

and the calendar one. Should it be `calendar:attach` no? With `calendar.attach` I get some nil reference inside the function.

Regardless, https://github.com/copycat-killer/lain/commit/e5e199c1488cd33b96e15f9879a4b9abcfdcc572 seems to be missing something.
The quick test I did resulted in calendar only showing up on the first screen. I'll do better tests later.
The fs widget displays on the correct screen but now the notification window doesn't vanish when unhovered. I need to click it to go away.

copycat-killer:
> The examples you mentioned didn't work for me

Sorry, I was in a hurry. Correct version:

```lua
myfs = lain.widgets.fs({
    notification_preset = { screen = mouse.screen }
})
```

and yes, it was `calendar:attach`. Just a typo.

> The quick test I did resulted in calendar only showing up on the first screen. 

Did you try this?

```lua
lain.widgets.calendar:attach(mytextclock, { followmouse = true })
```

> The fs widget displays on the correct screen but now the notification window doesn't vanish when unhovered. I need to click it to go away.

Another typo. Fixed.

Unode:
Ok, with the last fix the fs widget now works fine.

The calendar not yet. Still always shows on screen 1. Regardless of the position of the mouse. Using:
```
lain.widgets.calendar:attach(mytextclock, { fg = theme.fg_focus,
                                            font = theme.font_name,
                                            font_size = theme.font_size,
                                            followmouse = true })

```
Inspecting the code I think I found the issue.
The `local` keyword [here](https://github.com/copycat-killer/lain/blob/master/widgets/calendar.lua#L94) causes `scrp` to be `nil` after the end of the if clause.

copycat-killer:
I'm doing my worst lately.

Removed `local`. Tell me now.

Unode:
All good now.

Thanks and sorry for the trouble.

copycat-killer:
No, thanks to you for enhancing my work.

PR #132, opened 2015-08-23
fix naughty.notify render problem, when bg color with alpha is used

everslick:
fixed a copy paste error in color initialization on the go


everslick:
as requested

copycat-killer:
Thanks.

Goh, I'm so lazy in this period.

PR #133, opened 2015-08-29
Make weather forecast notification text customizable

d125q:
This pull request makes the weather forecast notification text customizable. The default behavior remains unchanged.

Now there is an additional argument, `notification_text_cmd`, which is a function taking the day (should we rename this internally as `date`?) as returned by `date_cmd`, the description, the minimum temperature, and the maximum temperature, and should return a string representing the weather forecast for that day. In order to preserve the previous behavior, the default function returns `string.format("<b>%s</b>: %s, %d - %d  ", day, desc, tmin, tmax)`.

### DIFF
--- a/widgets/weather.lua
+++ b/widgets/weather.lua
@@ -132,7 +132,7 @@ local function worker(args)
     weather.attach(weather.widget)
 
     newtimer("weather-" .. city_id, timeout, weather.update)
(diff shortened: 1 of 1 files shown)


copycat-killer:
Just one trivial thing: renamed `cmd` to `fun` because it actually a function.

Thanks for the rest.

Issue #114, opened 2015-07-07
question concerning timers in lain


ff2000:
helpers.newtimer seems to not use stored timers efficiently. If I add more widgets of the same type (e.g. there are several batteries or sound cards or network devices etc) it creates a new timer and overwrites the previous one in the table (as I undertand LUA - which might be wrong). vicious indexes the timer table with the actual timeout which really seems to greatly affect wakeups and general CPU usage (0-1% in contrast to 2-3% with lain, both running a comparable set of widgets)
Furthermore vicious seems to efficiently cache formatted strings (don't get that part of the code entirely ATM, will have to stare at it some more time ;))

I think lain could take a similar approach to maximize performance/energy efficiency.

copycat-killer:
You are right on `bat`, `alsa` and `alsabar`, but wrong on `net`: timer here is set with `iface`, which is unique.

I updated the following widgets in order to use unique ids for timer:

* `alsa`
* `alsabar`
* `bat`
* `tpbat`
* `fs`
* `weather`

but not for

* `cpu`
* `mem'
* `mpd`
* `sysload`
* `temp`

because their instance sets are always singletons.

Test 6457c4d and tell me.

ff2000:
Thx for the commits! I am away for a week and will test it when I'm back.
Though still this doesn't share timers like vicious: Setting the timeout for CPU, MEM, NET to 1 sec. will use only 1 timer not 3 like with lain. That way the awesome process wakes up only once. Don't you like that approach?

copycat-killer:
So, not just unique ids, but also the *same* timeout for one-instance-only widgets, right?

Ok, but keep in mind that a user can change timeouts arbitrarily, so the general complexity is not affected.

Can you tell me how do you precisely test this?

copycat-killer:
Plus, please note that `imap`, `mpd` and `weather` are asynchronous, so if you're also using them, of course your config is a little bit more cpu demanding.

copycat-killer:
@ff2000 Can you update me on this? Also, tell me how do you do your tests.

ff2000:
The comparison between vicious and lain probably was not that great. As I said I took a similar set of widgets and the same timeouts and compared the resulting CPU usage. Of course the widgets (and the logic behind them) is different. I would need to write my own widgets and adopt them to both APIs.
But I can at least get a vicious-like implementation in lain if I simply change lain/helpers.lua:
```diff
diff --git a/helpers.lua b/helpers.lua
[...]

PR #134, opened 2015-08-29
only update widget text if it has changed since the last timer tick

everslick:
I realized that widget.set_markup() is kinda bottleneck. On my atom netbook awesome has around 8% cpu load if idle, when i set the timer to 1 second for each of my widgets. 

this patch reduces cpu load to 4-5% while still maintaining high refresh rates. fine tuning the timers reduces the cpu load further (I set them to 1 for testing the patch).

pls note, I only changed the widgets i use, the others, I cannot really test. would be nice if at least the base widget could get enhanced the same way, but i failed to make a clean implementation. my Lua skills are rather poor.

### DIFF
--- a/helpers.lua
+++ b/helpers.lua
@@ -76,7 +76,6 @@ end
 
 helpers.timer_table = {}
 
---function helpers.newtimer(_name, timeout, fun, nostart)
 function helpers.newtimer(_name, timeout, fun, nostart)
     local name = timeout
 
--- a/helpers.lua
+++ b/helpers.lua
@@ -76,10 +76,17 @@ end
 
 helpers.timer_table = {}
 
-function helpers.newtimer(name, timeout, fun, nostart)
-    helpers.timer_table[name] = capi.timer({ timeout = timeout })
+--function helpers.newtimer(_name, timeout, fun, nostart)
+function helpers.newtimer(_name, timeout, fun, nostart)
+    local name = timeout
+
+    if not helpers.timer_table[name] then
+        helpers.timer_table[name] = capi.timer({ timeout = timeout })
+        helpers.timer_table[name]:start()
+    end
+
     helpers.timer_table[name]:connect_signal("timeout", fun)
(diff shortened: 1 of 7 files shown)


copycat-killer:
Thanks for your effort, I've never thought about this, "thanks" to my fast CPUs.

I'll see if I can write an helper to automatize the check, otherwise I'll merge this and do the rest.

>Fine tuning the timers reduces the cpu load further (I set them to 1 for testing the patch).

@ff2000 Can you try this patch with fine tuning and see if it solves #114?

everslick:
maybe a simple wrapper around set_markup() will do? my approach tries do push as much as possible into the 'if' to minimize cpu load. on the other hand we can avoid maybe even more cpu cycles on each timer tick by combining ff2000's approach with mine.

ff2000:
I'm waiting for https://github.com/awesomeWM/awesome/pull/404 - it might be many of those performance issues could be solved by that.
I also used qtile now for some days. Though it is in development for quite some years it still lacks too many things, but does extremely well performance-wise. It idles at ~1.5% CPU usage while I use widgets (cpu, net, memory graphs) that drive awesome to >10% here. I did not have the time to look at the code too deep (and I won't have it for the next weeks/months), could be inspiring for awesome, too.
Concerning this PR: Couldn't this be done in a more generic way? Implementing cahing for every widget seems to be quite some boilerplate. Vicious IMHO does caching quite good (easy to use and generic). Talking about timers in #114 I thought this was already rejected (resetting timers for individual widgets), and I can live with it.

I will try this PR the next days. Hope that's OK.

copycat-killer:
The new widget system looks amazing, but probably it'll take some time.

> I thought this was already rejected

No, I do want to improve performances and I'm happy you guys made me focus on it. It's just I have little time at the moment.

copycat-killer:
I have to reject this because violates #141 solution, has conflicts with master, and does not consider all the widgets.

But don't worry, I applied the patch anyway. Check #114 thread.

PR #136, opened 2015-09-06
Add a new layout: centerhwork, based on centerwork

Ganneff:
centerwork is mostly useful on wide screens, but if you happen to
have (one) screen turned 90° you have a pretty high, but slim, space
for windows. centerwork is not useful there.

centerhwork takes the same concept as centerwork, just makes the main
window expand horizontally not vertically, and the 4 additional windows
are put ontop/below the main window, thus using the huge vertical space
much better.

This isn't too much of a work by me, mostly just changing name and then fiddling with the math a bit, so if you care and want to remove the (C) line at the top that I added, feel free to. Its just useful for me, and it might be for others, hence the pull request. As usual: it worksforme[TM] but may eat your fridge, noone knows. :)


copycat-killer:
@InfinityTotality Could you please integrate this one with all the recent patches?

InfinityTotality:
I've added it to my repository with all the sizing fixes. I also didn't realize the top auxiliary windows in the original centerwork were a full useless_gap larger, so I added a fix for that to the original and this one. The commit is here: https://github.com/InfinityTotality/lain/commit/351ceea5ad56043a6c68c890247fbcf49ac03486

copycat-killer:
Thanks again. :+1:

Ganneff:
Thanks!

Issue #135, opened 2015-08-30
Master window not primary window in cascadetiled and centerwork layouts


InfinityTotality:
I'm not sure if this is intentional, but the master window is placed in the bottom right corner of the centerwork layout (using the default arrangement) and at the top of the group of cascaded windows in the cascadetiled layout. It would be nice if the master window was the main, large window in these layouts. In both of these cases, it seems that the window which is placed in the main area is the last window in the client list rather than the first. I am using the latest git version which I installed using the lain-git AUR package.

P.S. I went ahead and modified my layouts to work this way. I also rearranged the centerwork auxiliary windows to make a little more sense when navigating using the default prev/next client key bindings. All I did was change the index for the main window in cls to 1 instead of #cls and updated the for loop to arrange the remaining windows accordingly. For the aux window positions, I changed all the "if at ==" to be "if i - 2 ==" instead.

P.P.S. In addition to handling the clients in reverse order (using cls[#cls] instead of cls[1] as the master window), centerfair had a bunch of sizing/spacing issues. I think I got them all fixed (what was with all the hard-coded 2s for border compensation?), but you can check them in this commit: https://github.com/InfinityTotality/lain/commit/c91c2e2d4b7f2243c13a267f7bfe4f3df73b35e3. I made the changes on the assumption that the useless_gap distance should be measured from the outside of the window borders. Termfair and centerwork still have some issues, but uselesstile seems to have the spacing right at least.

P.P.P.S. Alright, I did a bunch of work, and I think I have all the sizing and spacing consistent across all the layouts. They all now properly adjust for client borders, even if all the clients have a different border size, and useless_gaps are applied consistently on all layouts (measured from the outside of the borders). I adjusted termfair to apply useless_gaps the same as all the other layouts, but I left cascade's borders as they were. It might make sense to apply the useless gaps around the edges there too. Let me know what you think of the changes: https://github.com/copycat-killer/lain/compare/master...InfinityTotality:master. There are still a couple 1 pixel discrepancies between uselesstile and the custom layouts due to rounding issues which I might try to fix, but I'm pretty happy with all the custom layouts.

PR #138, opened 2015-09-09
Many sizing fixes

InfinityTotality:
All the layouts had some minor issues with sizing and spacing of the clients, so I fixed a bunch of them and added accounting for individual client borders. I haven't touched uselessfair or uselessspiral, so they may have some 1 pixel discrepancies still, and uselesstile has the potential for being 1 pixel off when there are multiple slave columns, but those are the only ones I know about.

You may want to leave out the part where I changed the arrangement of the auxiliary windows in centerwork and the parts where I switched from using cls[#cls] to cls[1] for the master window depending on your preferences, but all the sizing and spacing changes can be applied without those.

### DIFF
--- a/layout/cascade.lua
+++ b/layout/cascade.lua
@@ -28,15 +28,12 @@ function cascade.arrange(p)
     local global_border = tonumber(beautiful.global_border_width) or 0
     if global_border < 0 then global_border = 0 end
 
-    -- Themes border width requires an offset.
-    local bw = tonumber(beautiful.border_width) or 0
-
     -- Screen.
     local wa = p.workarea
     local cls = p.clients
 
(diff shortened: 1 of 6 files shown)


copycat-killer:
First of all, thank you for the huge work. :dancers: 

Does this fix #135 as well?

InfinityTotality:
Yep, the original issue I was talking about in #135 was the part where centerfair and centerwork use cls[#cls] as the master window then tile cls[1] through cls[#cls - 1] as the slave windows. Awful.client.getmaster() returns cls[1], so I updated the layouts to tile that one as the master and tile cls[2] through cls[#cls] as the slave windows.

PR #142, opened 2015-09-22
Added name attribute to mpd widget

xentec:
Metadata from some extern sources (e.g. soundcloud) is not parsed properly to a ``title`` by mpd so the ``name`` attribute is set instead.
Helps to determine what is actually playing in ``settings()``.


xentec:
I just noticed that this PR was undone by 0b80b45bc9bca17f08e8055d364bac17f96129bd. Is this intended?

copycat-killer:
> Is this intended?

Nope.

PR #144, opened 2015-09-23
allow add_us_secondary to be turned off.

stwa:
problem with test expression and nil value.

### DIFF
--- a/widgets/contrib/kbdlayout.lua
+++ b/widgets/contrib/kbdlayout.lua
@@ -22,10 +22,12 @@ local function worker (args)
 
    local layouts          = args.layouts
    local settings         = args.settings or function () end
-   local add_us_secondary = args.add_us_secondary or true
+   local add_us_secondary = true
    local timeout          = args.timeout or 5
 
    local idx              = 1
+   
+   if args.add_us_secondary == false then add_us_secondary = false end
 
    -- Mouse bindings
    kbdlayout.widget:buttons(awful.util.table.join(


copycat-killer:
Why don't you pass `add_us_secondary = false` instead?

stwa:
Because it is not working, Lua is acting strange in this case.

Look at this example:
> print( 3>1 and 1 or "False" )        -- works
1
> print( 3>1 and false or "oops" )     -- failed, should return false
oops
> print( 3>1 and nil or "oops" )       -- failed, should return nil
oops

Example from here:
http://lua-users.org/wiki/ExpressionsTutorial

copycat-killer:
OK, I tried the widget and the problem is indeed present.

Anyway, that example is correct. Lua always returns the non boolean value, if it's not `nil`, and `nil` is interpreted as `false`.

So, in the second instance, you have `3>1 and false` which is `T and F = F`, so the result is `F or string`, which is converted to `nil or string` and returns `string`. The same applies for the third one.

stwa:
Thank you for your explanation, makes a lot more sense now!

Am 24. September 2015 11:51:04 MESZ, schrieb Luke Bonham <notifications@github.com>:
>OK, I tried the widget and the problem is indeed present.
>
>Anyway, that example is correct. Lua always returns the non boolean
>value, if it's not `nil`, and `nil` is interpreted as `false`.
>
>So, in the second instance, you have `3>1 and false` which is `T and F
>= F`, so the result is `F or string`, which is converted to `nil or
>string` and returns `string`. The same applies for the third one.
>
>---
>Reply to this email directly or view it on GitHub:
>https://github.com/copycat-killer/lain/pull/144#issuecomment-142877284

-- 
Diese Nachricht wurde von meinem Android-Mobiltelefon mit K-9 Mail gesendet.

PR #147, opened 2015-10-12
Fixed unaccessible bat object in bat.lua.

stwa:
bat is local to worker, therefore cannot be accessed from outside.


enkoder:
This fixed my error in bat.lua.

copycat-killer:
Right, I was still testing the big work done for #114.

Thanks.

stwa:
Thanks for your work, great to see that lain is still improving!

PR #148, opened 2015-11-02
weather.lua -- improve icon and notification text handling

d125q:
This pull request:

1. Makes the "N/A text" customizable;
2. Makes the textbox widget start off with the "N/A text" rather than empty;
3. Makes the icon start off with the "N/A icon" rather than empty;
4. Sets the N/A icon when the request to retrieve the weather, **not** the forecast, fails -- as the icon is tied to the current weather rather than the forecast.

If there's a point you don't agree with, or something wrong with the code, let me know.

### DIFF
--- a/widgets/weather.lua
+++ b/widgets/weather.lua
@@ -50,7 +50,7 @@ local function worker(args)
     local followmouse           = args.followmouse or false
     local settings              = args.settings or function() end
 
-    weather.widget    = wibox.widget.textbox(weather_na_text)
+    weather.widget    = wibox.widget.textbox(weather_na_markup)
     weather.icon_path = icons_path .. "na.png"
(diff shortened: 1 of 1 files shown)


copycat-killer:
It's ok for me.

Speaking of wich, I was thinking to switch to [erikflowers/weather-icons](https://github.com/erikflowers/weather-icons). They seem prettier than the current ones, don't they?

d125q:
@copycat-killer 
That font looks awesome.  It is probably not too hard to get PNGs from there, but the coloring might be tricky.  For example, my entire theme is in solarized colors, and I can color the icons the same as the text, but I'm not sure how general that solution would be.

copycat-killer:
Who talked about PNGs? I'll grab the font itself. :dancers:

Issue #158, opened 2016-01-01
[solved] Battery percentage can be more than 100%


JJK96:
The bat widget shows percentages up to 121%. I looked at the code for the widget and saw that there is a catch for that, so this should not be possible. 

Also I suggest to use the /sys/class/power_supply/BAT0/capacity file to get the capacity directly instead of using three files and then working out the capacity manually.

aajjbb:
I can confirm that it happens with me as well. I [pulled](https://github.com/copycat-killer/lain/pull/159) a small patch to fix it.

copycat-killer:
@JJK96 Check #160

aajjbb:
Can you reproduce this problem with the new patch @JJK96 ? I'm not able to reproduce it because it used to happen in an aleatory manner. I think #161 finally fixed. Maybe this issue can be closed now.

JJK96:
sorry for my late reaction, I just tested the new patch and it fixed the issue, it can be closed.

copycat-killer:
Don't rename it, you can close it yourself.

PR #159, opened 2016-01-02
Battery percentage going above 100% fix

aajjbb:
Hi,

This is a small patch to not allow battery percentage go above 100% as mentioned by @JJK96 in [issue](https://github.com/copycat-killer/lain/issues/158)


JJK96:
Asuming that your patch works (haven't tested it yet) the if then else statement that is around your code is useless then, maybe remove it?

aajjbb:
Actually I just realized I haven't done nothing useful here. I haven't paid attention to the other ifs in the block. Actually I supposed the issue is due to an rounding problem with the division. I don't know lua well enough to assure whether this is the issue, but it's the most probable one.

Please, ignore my silly patch.

JJK96:
I will try to debug it tomorrow, will post a patch if i found a fix

aajjbb:
I think I've found the real source of the issue. I'll post a patch soon.

PR #167, opened 2016-02-03
Added dynamic centerwork layout

antonsson:
I created a new layout based on centerwork but instead of leaving it hardcoded to 4 slaves this layout will fill the slave-columns regardless of how many slave-clients there is. The layout is based on original centerwork layout.

This is my first pull-request on github and also my first lua-code so I hope I didn't miss anything essential:)

### DIFF
--- a/layout/centerworkd.lua
+++ b/layout/centerworkd.lua
@@ -1,0 +1,123 @@
+
+--[[
+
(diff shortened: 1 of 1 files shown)


copycat-killer:
Good job. :dancers:

PR #175, opened 2016-03-01
Added amixer card selection support.

kewitz:
I've added support for picking the right card on `alsa` and `alsabar` widgets. Card number defaults to 0.

You can specify the card number using:
```lua
volumewidget = lain.widgets.alsa({
    card = "1",
    settings = function()
        header = " Vol "
        vlevel  = volume_now.level

        if volume_now.status == "off" then
            vlevel = vlevel .. "M "
        else
            vlevel = vlevel .. " "
        end

        widget:set_markup(markup(accent, header) .. vlevel)
    end
})
```

I've also added support for mouse-scrolling volume control and clicking mute/unmute on the `alsa` widget.

### DIFF
--- a/widgets/alsa.lua
+++ b/widgets/alsa.lua
@@ -1,15 +1,16 @@
 
 --[[
-                                                  
-     Licensed under GNU General Public License v2 
-      * (c) 2013, Luke Bonham                     
-      * (c) 2010, Adrian C. <anrxc@sysphere.org>  
-                                                  
+
+     Licensed under GNU General Public License v2
+      * (c) 2013, Luke Bonham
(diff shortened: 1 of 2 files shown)


copycat-killer:
You can set your card number with `cmd`: 
```lua
volumewidget = lain.widgets.alsa({
    cmd = "amixer -c 1",
    settings = function()
        header = " Vol "
        vlevel  = volume_now.level

        if volume_now.status == "off" then
            vlevel = vlevel .. "M "
        else
            vlevel = vlevel .. " "
        end

        widget:set_markup(markup(accent, header) .. vlevel)
    end
})
```
As for the mouse signals, they are user-dependent and burden the widget (I don't need them, for instance).

Issue #177, opened 2016-03-10
bat.lua - watt is being computed wrongly


andering:
Hello, 
correct me if i'm wrong 
but it seems to me that with widget bat.lua and his property watt (current Watt usage?), 
is being computes as **W** = **I** * **V** , so W = power_now * voltage_now ,
BUT power_now doesn't contain **I** , but already current usage, so **W**

```
local rate  = first_line(bstr .. "/power_now") or first_line(bstr .. "/current_now")
local ratev = first_line(bstr .. "/voltage_now")

bat_now.watt = string.format("%.2fW", (rate * ratev) / 1e12)
```

Thank you for making it clear for me.

aajjbb:
It makes sense. 

If _power_now_ is not available, then usually _current_now is, the fact is that _power_now_ is already measured in _uW_, but _current_now_ is in _uA_, so it has to be multiplied by _voltage_now_ to become in fact power in watts.

stwa:
But what **if** `power_now` is available? It looks like **W = W * V** (according to the code snippet)

andering:
Right, that was my point. If power_now is present, W is calculated wrongly.

copycat-killer:
So if `power_now`, we take `rate` directly, otherwise we do the computation. Did I understand right?

@andering @stwa Please tell me if latest commit is ok.

stwa:
See my commit, suggestion to calculate watt for battery.
Not sure about the time_rat, was it correct before?

copycat-killer:
Your pull request was perfectly equivalent to my commit.

But I merged it anyway because it avoids to use a flag variable and so it uses less lines.

`time_rate` should be correct, as far as I can see.

@andering Give us your opinion.

andering:
Yes now it looks OK,

time_rat should be ok in case we can count that we have one or both from these two options (energy_now, energy_full, power_now) or (charge_now, charge_full,current_now) because they use same units, **if they for some reason mix, it would come out with various results** :)

Only issue now is this
 
`bat_now.watt = string.format("%.2fW", ratep)`

units are in uW so it needs to be divided by 1e6 to get Watts

`bat_now.watt = string.format("%.2fW", ratep / 1e6)`

my version of module (more explanatory) is here [bat.lua](https://github.com/andering/lain/blob/master/widgets/bat.lua
) , but i removed things which i saw as unreasonable like checking for negative values.

copycat-killer:
Please pull me your `bat.lua` as well.

[...]

PR #183, opened 2016-03-23
#177 bat.lua fix

andering:
It's my fir pull so be merciful with me :).
Looks like mess because i use tabulators not spaces. What should i do with that ?

### DIFF
--- a/widgets/bat.lua
+++ b/widgets/bat.lua
@@ -1,10 +1,10 @@
 
 --[[
-                                                  
-     Licensed under GNU General Public License v2 
-      * (c) 2013,      Luke Bonham                
-      * (c) 2010-2012, Peter Hofmann              
-                                                  
+												  
+	 Licensed under GNU General Public License v2 
+	  * (c) 2013,      Luke Bonham                
+	  * (c) 2010-2012, Peter Hofmann              
+												  
 --]]
 
 local newtimer     = require("lain.helpers").newtimer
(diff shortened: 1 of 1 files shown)


andering:
Ok now i can see what happends when all those >0 conditions are not there.
When i unplug and plug battery charger power_now is for moment at 0, so in
`rate_time = (energy_full - energy_now) / rate_power or rate_current`
we have division by 0 so infinite time for charging.

But to just limit output time between 0 - 23 is just doesn't seems right.
First i would get rid of division by 0 or maybe skip the round or signalize ?

copycat-killer:
> First i would get rid of division by 0 or maybe skip the round?

Timeout is 30 seconds, so we can just skip the rest of the function if rate_power is zero.

andering:
```
-- if rate = 0 or rate not defined skip the round
if not (rate_power and rate_power > 0) and 
	not (rate_current and  rate_current > 0) 
then
	return 
end
```

So while im newbie in lua i hope simple return is fine. Tested on my Desktop and works fine.

andering:
So, now should be ok.

```
-- if rate = 0 or rate not defined skip the round
if 	not (rate_power and rate_power > 0) and 
	not (rate_current and  rate_current > 0) and
	not bat_now.status == "Full"
then
	return 
end
```

copycat-killer:
Seems ok to me too. Thanks for the work.

Issue #182, opened 2016-03-23
cpu.lua - multicore


andering:
Hello,
i updated cpu.lua with multicore support for my own purposes,
should i pull it ? :)  [cpu.lua](https://github.com/andering/lain/blob/master/widgets/cpu.lua)

copycat-killer:
Go for it. :dancers:

andering:
So i pull it #183

andering:
#184

PR #184, opened 2016-03-23
#182 - Added multicore functionality into cpu.lua

andering:


copycat-killer:
Not merged because of tabs, but implemented in 14cab4e.

Thanks.

PR #187, opened 2016-04-02
crash fix - sink value initialized

MParolari:
Without particular input (no `scallback` or `settings` functions), `pulseaudio.sink` isn't defined and we can't use it in `string.format("pulseaudio-%s", pulseaudio.sink)` when timer is created (`newtimer` function).
Alternatively, we could remove `string.format` function and let sink value undefined.

I'm working on a better solution, but this fix should work and prevent crashing when `lain.widgets.pulseaudio()` is called in rc.lua without arguments.


copycat-killer:
There's a very simple solution.

Since every `lain.helpers.newtimer` is instantiated only once, we can attach to a single timer multiple signals. Using `args.timeout` instead of `pulseaudio.sink` (something I forgot to remove) will be enough.

MParolari:
Ok, I don't know how `newtimer` works, I thought that the "timer name" had to be unique....
Thanks

PR #194, opened 2016-04-16
Fix calendar highlighting by letting user handle it

bradst:
This fixes a calendar highlighting bug that I noticed earlier today (April 16):

lain-calendar

I use `cal -w` to show the week numbers, and the week number 16 was incorrectly highlighted instead of the day. This patch fixes the bug by letting `cal` itself do the highlighting by passing `--color=always` and replacing the reverse-video escape codes with <span> markup. This is how it looks with the patch:

lain-calendar-fixed

I tested all the command-line options for `cal` on Arch Linux, and the only escape codes it emits are for reverse video, so this won't lead to any garbage characters getting through unless there is a non-standard `cal` installed. Calendar widgets created with a user-defined command will need to be updated to include `--color=always`.

### DIFF
--- a/widgets/calendar.lua
+++ b/widgets/calendar.lua
@@ -50,8 +50,8 @@ function calendar:show(t_out, inc_offset, scr)
         calendar.notify_icon = calendar.icons .. today .. ".png"
 
         -- bg and fg inverted to highlight today
-        f = io.popen(string.format("%s | sed -r -e 's/_\\x08//g' -e '0,/(^| )%d($| )/ s/(^| )%d($| )/\\1<b><span foreground=\"%s\" background=\"%s\">%d<\\/span><\\/b>\\2/'",
-                     calendar.cal, today, today, calendar.bg, calendar.fg, today))
+        f = io.popen(string.format("%s | sed -e 's|\x1b\\[7m|<span foreground=\"%s\" background=\"%s\">|g' -e 's|\x1b\\[27m|</span>|g'",
+                     calendar.cal, calendar.bg, calendar.fg))
 
     else -- no current month showing, no day to highlight
        local month = tonumber(os.date('%m'))
@@ -101,7 +101,7 @@ end
 function calendar:attach(widget, args)
     local args = args or {}
 
-    calendar.cal         = args.cal or "/usr/bin/cal"
+    calendar.cal         = args.cal or "/usr/bin/cal --color=always"
     calendar.icons       = args.icons or icons_dir .. "cal/white/"
     calendar.font        = args.font or beautiful.font:gsub(" %d.*", "")
     calendar.font_size   = tonumber(args.font_size) or 11


copycat-killer:
> unless there is a non-standard `cal` installed

That's the reason why I decided to implement the highlighting.

On Arch, `cal` it's the original Unix one as it appears in Linux kernel utils, but on other distros is reimplemented differently (i.e. some flags diverge).

If I accept the pull request, every non Arch user should redefine `args.cal` according to its distro implementation. I dont' like this very much.

In GNU `sed`, you can append a numeric flag at the end of an expression to define the occurrence on which operate. I can add such flag to `sed` and make it user-definable.

This way we won't touch anything, but you should be able to have weeks in your popup.

Try this patch:

```patch
diff --git a/widgets/calendar.lua b/widgets/calendar.lua
index bbd6eed..fd80501 100644
--- a/widgets/calendar.lua
+++ b/widgets/calendar.lua
@@ -50,8 +50,8 @@ function calendar:show(t_out, inc_offset, scr)
         calendar.notify_icon = calendar.icons .. today .. ".png"
 
         -- bg and fg inverted to highlight today
-        f = io.popen(string.format("%s | sed -r -e 's/_\\x08//g' -e '0,/(^| )%d($| )/ s/(^| )%d($| )/\\1<b><span foreground=\"%s\" background=\"%s\">%d<\\/span><\\/b>\\2/'",
-                     calendar.cal, today, today, calendar.bg, calendar.fg, today))
+        f = io.popen(string.format("%s | sed -r -e 's/_\\x08//g' -e '0,/(^| )%d($| )/ s/(^| )%d($| )/\\1<b><span foreground=\"%s\" background=\"%s\">%d<\\/span><\\/b>\\2/%s'",
+                     calendar.cal, today, today, calendar.bg, calendar.fg, today, calendar.today_occ()))
 
     else -- no current month showing, no day to highlight
        local month = tonumber(os.date('%m'))
@@ -102,6 +102,7 @@ function calendar:attach(widget, args)
     local args = args or {}
 
     calendar.cal         = args.cal or "/usr/bin/cal"
+    calendar.today_occ   = args.today_occ or function() return 1 end
     calendar.icons       = args.icons or icons_dir .. "cal/white/"
     calendar.font        = args.font or beautiful.font:gsub(" %d.*", "")
     calendar.font_size   = tonumber(args.font_size) or 11
```

And then make `args.today_occ` outputs the occurrence of today number you want to highlight.
For instance:

```lua
-- Calendar
lain.widgets.calendar:attach(mytextclock, {
    cal = "/usr/bin/cal -w",
    today_occ = function()
        if os.date("%d") == os.date("%W") then
            return 2
        else
            return 1
        end
    end
})
```

The patch should also work if somehow there are more than 2 occurrences of today number in your `cal` output.

copycat-killer:
@bradst Did you try the patch?

bradst:
@copycat-killer Sorry for the late response. Your patch indeed solves the issue, although it seems a bit inflexible. What would you think about exposing the `string.format` call as an attribute instead? For example:

```
calendar.cal_format = args.cal_format or function()
    return string.format("%s | sed -r -e 's/_\\x08//g' -e '0,/(^| )%d($| )/ s/(^| )%d($| )/\\1<b><span foreground=\"%s\" background=\"%s\">%d<\\/span><\\/b>\\2/'", calendar.cal, today, today, calendar.bg, calendar.fg, today))
end

```

Issue #191, opened 2016-04-03
Net widget picks up the first configured device


Unode:
Hi Luke,

I started using docker on my machine and the net widget was no longer displaying any activity even when I was sure I was transferring something.

After inspecting the code I realized that if no interface is specified the widget picks up the first one "UP" where "first" means alphabetical order.
Before installing docker this worked fine as either the ethernet or wifi interfaces were picked up, however after installing docker, `docker0` is always the first interface to match. Since I'm not doing any transference inside docker the counters are at zero most of the time.

On the other hand, on the laptop I often toggle between cable and wifi as I move around and sometimes both are "UP" simultaneously.
Setting up a fixed interface on the configuration doesn't work that well on this scenario so either I have one widget per interface or I won't get the right reading.

So after all this I started to wonder if it would be possible to add up the transference rates of all the interfaces that are "UP". Optionally limiting to a few interfaces by passing a table or iterator via `args.iface`.
The current "string" interface would still be kept for backwards compatibility.

Would you be ok with this functionality?

copycat-killer:
Notice that `net.get_device` can be replaced by a function in `args.iface`. But if you're considering to enhance the widget by automatically detecting an often changing `iface`, then go on.

You could extend `net.get_device` by passing it an argument, i.e. "static"/"dynamic". If "static", it'll work like usual, otherwise, it'll do your new functionality.

PR #196, opened 2016-05-08
Support for multiple interfaces on net widget

Unode:
Addresses issue #191 by accepting an array if interface names via `args.iface`.

Compatibility with old API was kept by detecting if a string was passed.

Further testing with multiple devices is welcome.

### DIFF
--- a/widgets/net.lua
+++ b/widgets/net.lua
@@ -21,13 +21,13 @@ local setmetatable = setmetatable
 -- lain.widgets.net
 
 local function worker(args)
-    local net = { last_t = 0, last_r = 0 }
+    local net = { last_t = 0, last_r = 0, devices = {} }
 
-    function net.get_device()
+    function net.get_first_device()
         local ws = helpers.read_pipe("ip link show | cut -d' ' -f2,9")
         ws = ws:match("%w+: UP") or ws:match("ppp%w+: UNKNOWN")
-        if ws  then return ws:match("(%w+):")
-        else return "network off" end
+        if ws then return { ws:match("(%w+):") }
+        else return {} end
     end
 
     local args     = args or {}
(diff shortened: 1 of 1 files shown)


copycat-killer:
There are some issues:
* Lines [56-58](https://github.com/Unode/lain/blob/multiple-net-widget/widgets/net.lua#L56-58) seem redundant, if you already do the retrocompatibility check in [43-48](https://github.com/Unode/lain/blob/multiple-net-widget/widgets/net.lua#L43-48)
* Why do you use `net.last_t` and `net_last_r` in lines [82](https://github.com/Unode/lain/blob/multiple-net-widget/widgets/net.lua#L82)-[85](https://github.com/Unode/lain/blob/multiple-net-widget/widgets/net.lua#L85)? If I test using, for instance, `iface = { "wlan0", "wlan0" }`, I got negative values; If I use `dev_before` instead of `net`, I got the right values (i.e. the same obtained with the single device)

* Empty line [96](https://github.com/Unode/lain/blob/multiple-net-widget/widgets/net.lua#L96) should be removed
* Wiki needs to be updated

Check the following counter-patch and test if it works.

```diff
--- net.lua	2016-05-08 17:36:58.193173372 +0200
+++ net.new.lua	2016-05-08 17:30:28.690438985 +0200
@@ -26,7 +26,7 @@
     function net.get_first_device()
         local ws = helpers.read_pipe("ip link show | cut -d' ' -f2,9")
         ws = ws:match("%w+: UP") or ws:match("ppp%w+: UNKNOWN")
-        if ws  then return {ws:match("(%w+):")}
+        if ws then return { ws:match("(%w+):") }
         else return {} end
     end
 
@@ -47,20 +47,15 @@
         iftable = iface
     end
 
-    -- mark all devices as initially online/active
+    -- Mark all devices as initially online/active
     for i, dev in ipairs(iftable) do
         helpers.set_map(dev, true)
     end
 
     function update()
-        if next(iftable) == nil then
-            iftable = net.get_first_device()
-        end
-
         -- These are the totals over all specified interfaces
         net_now = {
-            -- Bytes since last iteration
-            sent = 0,
+            sent     = 0,
             received = 0
         }
 
@@ -70,7 +65,7 @@
 
         for i, dev in ipairs(iftable) do
             local dev_now = {}
-            local dev_before = net_now[dev] or { last_t = 0, last_r = 0 }
+            local dev_before = net_now[dev] or net
 
             dev_now.carrier  = helpers.first_line(string.format('/sys/class/net/%s/carrier', dev)) or '0'
             dev_now.state    = helpers.first_line(string.format('/sys/class/net/%s/operstate', dev)) or 'down'
@@ -79,10 +74,10 @@
             local now_r      = tonumber(helpers.first_line(string.format('/sys/class/net/%s/statistics/rx_bytes', dev)) or 0)
 
             if now_t ~= dev_before.last_t or now_r ~= dev_before.last_r then
-                dev_now.sent     = (now_t - net.last_t) / timeout / units
+                dev_now.sent     = (now_t - (dev_before.last_t or 0)) / timeout / units
                 net_now.sent     = net_now.sent + dev_now.sent
                 dev_now.sent     = string.gsub(string.format('%.1f', dev_now.sent), ',', '.')
-                dev_now.received = (now_r - net.last_r) / timeout / units
+                dev_now.received = (now_r - (dev_before.last_r or 0)) / timeout / units
                 net_now.received = net_now.received + dev_now.received
                 dev_now.received = string.gsub(string.format('%.1f', dev_now.received), ',', '.')
             end
@@ -92,19 +87,14 @@
 
             net_now[dev] = dev_now
 
-            if not string.match(dev_now.carrier, "1") and notify == "on"
-
-            then
-                if helpers.get_map(dev)
-                then
-                    naughty.notify({
-                        title    = dev,
-                        text     = "no carrier",
-                        icon     = helpers.icons_dir .. "no_net.png",
-                        screen   = screen
-                    })
-                    helpers.set_map(dev, false)
-                end
+            if string.match(dev_now.carrier, "0") and notify == "on" and helpers.get_map(dev) then
+                naughty.notify({
+                    title    = dev,
+                    text     = "no carrier",
+                    icon     = helpers.icons_dir .. "no_net.png",
+                    screen   = screen
+                })
+                helpers.set_map(dev, false)
             else
                 helpers.set_map(dev, true)
             end
```

Unode:
Hi Luke,

Thanks for the feedback.

Without lines [56-58](https://github.com/Unode/lain/blob/multiple-net-widget/widgets/net.lua#L56-58) if no interface is requested and no network is up when awesome starts, no device would be considered and the widget would be running on `nil` the whole time. The way it is now it gets defined once a device is up (i.e. array no longer empty). I'll add a comment to clarify the intention.

Lines 82-85 is a typo. Your patch is correct.

`local dev_before = net_now[dev] or net` is not correct. If we do that and we have more than one device they will share the same array instead of a copy of it. The idea is to have one set of `last_t last_r` per device and one global for the total (sum of all devices).

I don't see the need for `(dev_before.last_t or 0)` if `local dev_before = net_now[dev] or { last_t = 0, last_r = 0 }` is kept. I can't see how `dev_before.last_t` wouldn't be a number.

Using `{ "wlan0", "wlan0" }` never crossed my mind I can't see why anyone would do that. Still, it was helpful to see the problem you mentioned and a few others related with preserving the global state.

You seem to have accidentally merged the pull request in one of your last commits (84313cc876bce8b346ca83934de57b50488ab8cd).

Updated pull request with patch suggestion and additional fixes.

Unode:
Missed a few details that were corrected in the last pushes.

I still don't know how to deal with the wiki properly so here's [a patch instead](https://gist.github.com/Unode/2e429a162b98ade49edd269bea1f4392).

copycat-killer:
> I don't see the need for (dev_before.last_t or 0) if local dev_before = net_now[dev] or { last_t = 0, last_r = 0 } is kept. I can't see how dev_before.last_t wouldn't be a number.

That was due to `net` being nil sometimes, because of some synch discrepancies.

> Using { "wlan0", "wlan0" } never crossed my mind I can't see why anyone would do that. Still, it was helpful to see the problem you mentioned and a few others related with preserving the global state.

Still, is the only use case I could use, and in if your first commit was right, it should have worked.

> You seem to have accidentally merged the pull request in one of your last commits (84313cc).

I know, I was planning to fix it after reading your reply.

copycat-killer:
I merged locally solving the commit conflict.

Thanks for your efforts.

PR #201, opened 2016-05-28
Support of multiple batteries

MaximeCheramy:
This patch adds the support of multiple batteries.

On laptops containing two batteries, the widget could eventually be used twice to watch each battery, but the warning popups should be common between all the batteries.

A list of batteries can now be given instead of a single one. Tested on a Lenovo T460S.


Exapiper:
Thanks for adding this...I had been really procrastinating trying to figure this out haha...

PR #204, opened 2016-06-08
mpd: make notifications hideable in widget config

rohieb:
I personally find notifications on every song change too disturbing. So I only want to add the widget to the bar, which shows the current song info, and I can look there if it interests me.

### DIFF
--- a/widgets/mpd.lua
+++ b/widgets/mpd.lua
@@ -37,6 +37,7 @@ local function worker(args)
     local music_dir   = args.music_dir or os.getenv("HOME") .. "/Music"
     local cover_size  = args.cover_size or 100
     local default_art = args.default_art or ""
+    local show_notifications = args.show_notifications == true
     local followmouse = args.followmouse or false
     local echo_cmd    = args.echo_cmd or "echo"
     local settings    = args.settings or function() end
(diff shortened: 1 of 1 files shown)


widgets/mpd.lua
```diff
@@ -37,6 +37,7 @@ local function worker(args)
     local music_dir   = args.music_dir or os.getenv("HOME") .. "/Music"
     local cover_size  = args.cover_size or 100
     local default_art = args.default_art or ""
+    local show_notifications = args.show_notifications == true
```
rohieb: Not the same pattern like the next line because `args.show_notifications or true` would still be evaluated to `true` if I set `show_notifications = false` in the input table for the widget.

copycat-killer:
The default behaviour is to show the notification, so I reverted your change. You can hide the popups with `args.notify = false`.

copycat-killer:
The default behaviour is to show the notification, so I reverted your change. You can hide the popups with `args.notify = "off"` (just like battery widget).

copycat-killer:
The default behaviour is to show the notification, so I reverted your change. You can hide the popups with `args.notify = "off"` (just like battery widget).

rohieb:
Yes, right, I did not think far enough and accidentally changed the default value. Thank you for the fix!

PR #205, opened 2016-06-20
MPD: add playlist position and mode flags

rohieb:
This adds the following information to the `mpd_now` table:

- pls_len (length of playlist)
- pls_pos (current position in playlist)
- single_mode (`true` if single mode is switched on, else `false`)
- repeat_mode (`true` if repeat mode is switched on, else `false`)
- random_mode (`true` if random mode is switched on, else `false`)
- consume_mode (`true` if consume mode is switched on, else `false`)


rohieb:
In action (including widget icons) on https://github.com/rohieb/dotfiles/commit/22fbb2d

PR #206, opened 2016-06-22
Added gpmpd widget

Kirogo:
Hello,
In a similar fashion as the mpd widget, I made a plugin for [Google Play Music Desktop Player](http://www.googleplaymusicdesktopplayer.com/).
I also have a wiki page ready, but as it seems to be a different git repo, so I didn't include it on this PR.

Thanks you for taking the time to maintaining lain !
Cheers,
Alexandre.

### DIFF
--- a/widgets/contrib/gpmdp.lua
+++ b/widgets/contrib/gpmdp.lua
@@ -1,0 +1,105 @@
+
+--[[
+
+        Licensed under GNU General Public License v2
+        * (c) 2016, Alexandre Terrien
+
+--]]
+
+local helpers       = require("lain.helpers")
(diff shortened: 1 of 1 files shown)


copycat-killer:
Ok, I still need (at least a link to) the wiki page though, then I'll merge.

Kirogo:
Thanks for merging this, and sorry for not having sent to you the wiki page I had done. I just noticed a little mistake on the wiki page you've done, the default value for notify is "off", because gpmdp already sends notifications. However, as this option can be deactivated on the player, I've kept it.

Cheers,
Alexandre.

copycat-killer:
Right, I've set it to "on".

PR #220, opened 2016-08-05
asyncshell: minor fix to spawn with shell invocation

setrofim:
Fixed an erroneous call to "awful.spawn.with_shell".

### DIFF
--- a/asyncshell.lua
+++ b/asyncshell.lua
@@ -52,7 +52,7 @@ function asyncshell.request(command, callback, timeout)
         id, formatted_command
     )
 
-    if type(awful.spawn) == 'table' then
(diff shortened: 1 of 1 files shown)


copycat-killer:
With your pull request, then what would be the point of that if-else statement?

Which issue is causing you that if-else?

setrofim:
When requiring tyrannical, I was getting a bunch of errors "attempt to call field 'with_shell' (a nil value)" on that line. But, yes, my fix is probably not correct; maybe the if condition needs to change

setrofim:
tg

setrofim:
I've updated the pull request to change the if condition instead.

PR #226, opened 2016-09-03
widget.base: Fix cached output

frizinak:
chicken or the egg syndrome, output is never declared, thus comparison always fails, so output is never assigned to the output of read_pipe, which results in the comparison failing which in turn ...


copycat-killer:
You're right, a silly typo sitting there for 6 months.

PS: at first I thought you were a bot.

frizinak:
Damn I apparently pulled 6 months+ worth of changes then ;)

A bot ^^ how so?

PR #234, opened 2016-10-25
Add Option to show pulseaudio sink in widget.

Doron-Behar:
Pull request for WiKi page is comming.

### DIFF
--- a/widgets/pulseaudio.lua
+++ b/widgets/pulseaudio.lua
@@ -26,16 +26,19 @@ local function worker(args)
    local scallback   = args.scallback
 
    pulseaudio.cmd    = args.cmd or string.format("pacmd list-sinks | sed -n -e '0,/*/d' -e '/base volume/d' -e '/volume:/p' -e '/muted:/p'")
+   pulseaudio.sink_cmd = args.sink_cmd or string.format("pacmd list-sinks | gawk \'{\nif ($0 ~ /\\*\\ index/) {\nwhile ($0 !~ /device\\.string/) {\ngetline\n}\nprint gensub(/\\\"/,\"\",\"g\",$3)\nnext\n}\n}\'")
    pulseaudio.widget = wibox.widget.textbox('')
 
    function pulseaudio.update()
       if scallback then pulseaudio.cmd = scallback() end
       local s = read_pipe(pulseaudio.cmd)
+      local sink = read_pipe(pulseaudio.sink_cmd)
 
       volume_now = {}
       volume_now.left  = tonumber(string.match(s, ":.-(%d+)%%"))
       volume_now.right = tonumber(string.match(s, ":.-(%d+)%%"))
       volume_now.muted = string.match(s, "muted: (%S+)")
+      volume_now.sink = sink
 
       widget = pulseaudio.widget
       settings()


copycat-killer:
The default sink is already defined in `pulseaudio.cmd`, instead of that long command, you can just add a simple `string.match` variable. 

But what is the purpose? The user should already know which the default sink is, or he can know with just a quick shell command.

Moreover, you can define `pulseaudio.scallback`, if your sink changes.

Doron-Behar:
You are right, after going a little bit deeper into that lua code, I've figured out how the `pulseaudio.cmd` works (as a shell command) and I realized that the long command I added is really not necessarily. (I copied it from a shell script I wrote once but your command is much simpler.. (I'm not very experienced with `sed` yet :[ )).

### BTW:

- I noticed that `volume_now.right` and `volume_now.left` both get the same value with `string.match`<sup>[ref](https://github.com/copycat-killer/lain/blob/master/widgets/pulseaudio.lua#L36)</sup>. Is this intentional? What do you think about the following changes? I tested it and it works.
```
       volume_now = {}
-      volume_now.left  = tonumber(string.match(s, ":.-(%d+)%%"))
-      volume_now.right = tonumber(string.match(s, ":.-(%d+)%%"))
+      volume_now.left  = tonumber(string.match(s, "left:.-(%d+)%%"))
+      volume_now.right = tonumber(string.match(s, "right:.-(%d+)%%"))
```

- I still think that printing the **name** of the current sink might be useful for some people, especially if you switch often the sinks like me with a keyboard shortcut. I still want to open a better PR which I hope will satisfy you. In addition **I'll edit the wiki only if you will accept the new PR** (It was the 1st time I edited a wiki page so I didn't know that It's not like a repository and that my changes will instantly be saved without your permission :| ). 

- I ran into an issue with pulseaudio and awesome one day that caused unnecessary errors. The problem was that pulseaudio didn't start from some reason with the awesome session. The widget kept printing unstoppable errors saying that `s` is `nil` and therefore the interpreter can't run `string.match()` for it. It can be fixed by add ` or ""` for every string in `volume_now`.

Tomorrow I'll put it all into a new PR and I'll be happy to see it merged with/without additional edits from you.

copycat-killer:
Tell me if you like 79687ba. Read the wiki for [updates](https://github.com/copycat-killer/lain/wiki/pulseaudio).

copycat-killer:
> - I noticed that `volume_now.right` and `volume_now.left` both get the same value with `string.match`<sup>[ref](https://github.com/copycat-killer/lain/blob/master/widgets/pulseaudio.lua#L36)</sup>. Is this intentional?

Using `left-right` won't work, because in some systems there's `0-1` instead. I added a loop to fetch every sink channel into a table, indexing the entries with numbers so to be machine independent.

I also fixed the `nil` issue for when pulseaudio is missing, and added sink `index`.

Doron-Behar:
1. It's great, I really liked you used `or "N/A"`. Still, what if someone wants to show the **name** of the sink and not only the **index** of it? How about adding to `pulseaudio.cmd` something like: `-e '/device\.string/p'` at the end?
Here's pretty much what I wanted to show in my PR:
```
-   pulseaudio.cmd    = args.cmd or string.format("pacmd list-sinks | sed -n -e '0,/*/d' -e '/base volume/d' -e '/volume:/p' -e '/muted:/p'")
+   pulseaudio.cmd    = args.cmd or string.format("pacmd list-sinks | sed -n -e '0,/*/d' -e '/base volume/d' -e '/volume:/p' -e '/muted:/p' -e '/device\\.string/p'")

-      volume.now.index = string.match(s, "index: (%S+)") or "N/A"
+      volume_now.index = string.match(s, "index: (%S+)") or "N/A"

+      volume_now.sink = string.match(s, "device.string = \"(%S+)\"") or "N/A"
 
       widget = pulseaudio.widget
       settings()
```
BTW:
There's a small typo I fixed in this diff: You wrote `volume.now.index` instead of `volume_now.index`

copycat-killer:
Done.

As I don't use Pulseaudio, please give me a sample output of:

```shell
pacmd list-sinks | sed -n -e '0,/*/d' -e '/base volume/d' -e '/volume:/p' -e '/muted:/p' -e '/device\\.string/p'
```
so that I can add it to the wiki.

Doron-Behar:
Here you go:
```
$ pacmd list-sinks | sed -n -e '0,/*/d' -e '/base volume/d' -e '/volume:/p' -e '/muted:/p' -e '/device\.string/p'
	volume: front-left: 18340 /  28% / -33.18 dB,   front-right: 18340 /  28% / -33.18 dB
	muted: no
		device.string = "front:1"
```
**Note the double backslash inserted in the lua code but not in the shell command**: 
In a shell you write:
```sh
(...) -e '/device\.string/p'
```
But inside the `string.format()` in the lua script you use:
```lua
(...) -e '/device\\.string/p'"
```

copycat-killer:
Yeah, I'm just lazy.

PR #235, opened 2016-11-01
Add applesmc widget for displaying fan speed on MacbookPro

nmquyet:
Need applesmc module to be enabled

### DIFF
--- a/widgets/contrib/applesmc.lua
+++ b/widgets/contrib/applesmc.lua
@@ -1,0 +1,51 @@
+local newtimer     = require("lain.helpers").newtimer
+
+local wibox        = require("wibox")
+
+local io           = { open = io.open }
+local tonumber     = tonumber
+
+local setmetatable = setmetatable
+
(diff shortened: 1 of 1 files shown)


copycat-killer:
Just use `base`. Start with something like this:

```lua
lain.widgets.base({
    cmd = "cat /sys/devices/platform/applesmc.768/fan1_input " ..
          "/sys/devices/platform/applesmc.768/fan2_input",
    settings = function()
        widget:set_text(string.gsub(output, "\n", " "))
    end
})
```

copycat-killer:
Just use `base`. Start with something like this:

```lua
applesmc = lain.widgets.base({
    cmd = "cat /sys/devices/platform/applesmc.768/fan1_input " ..
          "/sys/devices/platform/applesmc.768/fan2_input",
    settings = function()
        widget:set_text(string.gsub(output, "\n", " "))
    end
})
```

PR #239, opened 2016-12-04
Add fstype and exclude_fstype options 

2009:
These get passed to the dfs script and df to limit the display of the fs widget to specific filesystem types.

### DIFF
--- a/scripts/dfs
+++ b/scripts/dfs
@@ -11,10 +11,11 @@
 # -------------------------------------------------------------------------
 #   Decoding options
 # -------------------------------------------------------------------------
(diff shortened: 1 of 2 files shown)


copycat-killer:
I added the change to `dfs`, but implemented it differently in `fs`.

In order to set your options, just add a second argument (string) to `fs.show`. Example:

```lua
awful.key({ altkey }, "h", function () mypartition.show(5, "--exclude-type=tmpfs") end),
```

2009:
Awesome, thanks.

copycat-killer:
@2009 https://github.com/copycat-killer/lain/pull/259#issuecomment-271269309

PR #241, opened 2016-12-05
Fix typo in return statement

2009:


copycat-killer:
Damn, these typos.

2009:
Haha, that is why I should never commit code when I'm tired.

PR #246, opened 2016-12-11
Fixed fs widget showing error messages on mouseover

ikselven:
Everytime the mouse has been moved over the file system widget, an error message ("/usr/share/awesome/lib/naughty.lua:342: attempt to compare number with table") appeared instead of the file system info. Fixed by the attached changes.


copycat-killer:
Fixed by #247.

PR #259, opened 2017-01-09
Temporary file systems are not that informative for a fs usage popup.

wimstefan:


aajjbb:
It might not be that informative but I don't think it's the case to get rid of that. Maybe creating a new field in the constructor to let it up to the user on whether or not to show temporary file systems is more interesting.

copycat-killer:
As of [f5700be](https://github.com/copycat-killer/lain/commit/f5700be12b68f73d20e26a7293881707897d8f04), the `options` argument is moved from `fs:show` to constructor.

```lua
mypartition = lain.widgets.fs({ options = "--exclude-type=tmpfs" })
```

wimstefan:
Thank you so much! :smile:

Issue #261, opened 2017-01-11
[bat.lua] multiple batteries depends on the order of the list "batteries"


sim590:
I have the following code in my `rc.lua` file:
```lua
baticon = wibox.widget.imagebox(beautiful.widget_battery)
batwidget = lain.widgets.bat({
    batteries = {"BAT0", "BAT1"},
    settings = function()
        if bat_now.perc == "N/A" then
            widget:set_markup(" AC ")
            baticon:set_image(beautiful.widget_ac)
            return
        elseif tonumber(bat_now.perc) <= 5 then
            baticon:set_image(beautiful.widget_battery_empty)
        elseif tonumber(bat_now.perc) <= 15 then
            baticon:set_image(beautiful.widget_battery_low)
        else
            baticon:set_image(beautiful.widget_battery)
        end
        widget:set_markup(" " .. bat_now.perc .. "% ")
    end
})
```

upower outputs this:
```sh
$ upower --dump | grep perc
    percentage:          100%
    percentage:          25%
    percentage:          59.7501%
```

and the percentage displayed is 100%. Now, if I supply the batteries array in the following order instead `{"BAT1", "BAT0"}`, the percentage will be accurate, i.e 59%. The issue comes from this line:
```lua
    bat_now.status = bat_now.n_status[1]
```

I won't comment on this as there are no words to describe my thoughts. I'm uploading a patch for this and hope you can merge it soon. Thanks.

copycat-killer:
> I won't comment on this as there are no words to describe my thoughts.

Really?

I didn't assume that battery status is the first battery status. The variable `status` points to the first one for backwards compatibility with the old single battery API. Nothing less, nothing more.

Anyway, I recognize this may be convenient.

sim590:
> > I won't comment on this as there are no words to describe my thoughts.
>
> Really?

@copycat-killer. I appologize. My comment was intended to be "funny", but I recognize there's no way that on the internet someone that doesn't know me can expect this. In reality, it is not funny and I appologize.

Tell me if I should change something in my patch.

copycat-killer:
It's OK, no offence taken.

Thanks for the patch, I just have a single battery laptop, so I'd hardly thought about it.

PR #262, opened 2017-01-11
bat: redefine battery status with multiple batteries

sim590:
This addresses #261. 

Battery status was previously assumed to the first battery's status. This is wrong as it doesn't support the case where the first battery is full and the second is discharging, therefor showing battery percentage to 100% when the actual battery percentage is lower.

### DIFF
--- a/widgets/bat.lua
+++ b/widgets/bat.lua
@@ -108,7 +108,16 @@ local function worker(args)
             end
         end
 
+        -- When one of the battery is charging, others' status are either
+        -- "Full", "Unknown" or "Charging". When the laptop is not plugged in,
+        -- one or more of the batteries may be full, but only one battery
+        -- discharging suffices to set global status to "Discharging".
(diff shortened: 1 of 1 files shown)


widgets/bat.lua
```diff
@@ -1,10 +1,10 @@
 
 --[[
-												                        
-	 Licensed under GNU General Public License v2 
-	  * (c) 2013,      Luke Bonham                
-	  * (c) 2010-2012, Peter Hofmann              
-
```
sim590: @copycat-killer: Concerning writing style: this part was done automatically by my editor's config. Should I undo that? I tend to see spaces like this in other lua files.

copycat-killer: Yes, please. Those spaces are part of the signature.

sim590:
I think that the solution I wrote doesn't fit all cases. I'll provide another version of this when I get to work on this.

PR #266, opened 2017-01-12
fixing redshit spawn_with_shell to be compliant with awesome v4.0

aajjbb:
Old redshift used `awful.spawn_with_shell` which is now `awful.util.spawn_with_shell`


copycat-killer:
Actually, it's `awful.spawn.with_shell`.

aajjbb:
Fixed. Curious fact is that both functions exists.

PR #270, opened 2017-01-22
Generic shebang in scripts/dfs

Unode:
Some systems have bash at locations other than /usr/bin/bash
Relying on /usr/bin/env is the most compatible option.

### DIFF
--- a/scripts/dfs
+++ b/scripts/dfs
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env bash
 #
 #   Adapted from Eridan's "fs" (cleanup, enhancements and switch to bash/Linux)
 #   JM,  10/12/2004


copycat-killer:
I want to switch to a pure Lua implementation and get rid of `scripts/` definitely.

I'll start thinking about it once I finish making the module completely asynchronous and independent from `beautiful`. In the meantime, any help in that direction is well-accepted.

copycat-killer:
Plus, for the most compatible option, `/usr/bin/env sh` is maybe better.

Unode:
I don't know if the script is `dash` compatible.
The previous shebang was `bash` so I left it as is.
Didn't want to spend to much on it since I noticed other scripts were removed/replaced.

copycat-killer:
1. It is.
2. I'll change it to `/usr/bin/env sh` for the time being.
3. That's my point too, read my first answer.

PR #273, opened 2017-01-22
some sticky fun

zhrts:
I wanted a sticky quake terminal, but the module overrides awful.rules. This lets the quake application be sticky witihout changing the default behaviour.

### DIFF
--- a/util/quake.lua
+++ b/util/quake.lua
@@ -66,8 +66,8 @@ function quake:display()
     client.size_hints_honor = false
     client:geometry(self:compute_size())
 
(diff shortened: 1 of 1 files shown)


copycat-killer:
I added a `settings` function, which accepts the client as input and allows to customize its properties.

```lua
-- set the client sticky
lain.util.quake({ settings = function(c) c.sticky = true end })
```

Read [here](https://awesomewm.org/doc/api/classes/client.html#Object_properties) for the complete list of properties.

zhrts:
Well done, thank you 👌

PR #275, opened 2017-01-23
Alsabar.lua changes

ProfessorOptics:
Fixed spelling heigth = height; added support for mouse buttons 1 through 5; added togglechannel as per previous version using new api; small fixes to accomodate togglechannel variable.

Also removed trailing whitespaces; added terminal variable for local use; mixer variable to store "amixer"

any issues, let me know.

### DIFF
--- a/widgets/alsabar.lua
+++ b/widgets/alsabar.lua
@@ -1,10 +1,7 @@
-
 --[[
-                                                  
-     Licensed under GNU General Public License v2 
-      * (c) 2013, Luke Bonham                     
-      * (c) 2013, Rman                            
-                                                  
+     Licensed under GNU General Public License v2
+      * (c) 2013, Luke Bonham
+      * (c) 2013, Rman
 --]]
 
 local helpers      = require("lain.helpers")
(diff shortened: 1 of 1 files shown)


copycat-killer:
I decided to remove  the buttons stuff, but it seems people want them, so ok.

copycat-killer:
Wait, no, sorry, there is a specific motivation for those changes: removing `beautiful` and external variable dependencies, and let the user decide about buttons.

The reason is portability and modularity: you'll see this applied in the next `awesome-copycats` commits.

Now `notifications` is `notification_preset`, and if you want buttons just add the snippet in your `rc.lua`.

ProfessorOptics:
I wasn't aware of that, and now i understand why it was missing after the update. I'll keep that in mind and update my scripts accordingly.

ProfessorOptics:
Does adding toggle channel and its various functions remove the portability? I could modify the script according to your recent changes and send another pull request?

copycat-killer:
The point is, we don't need `togglechannel`, since `amixer` or `args.cmd` grabs it anyway.

ProfessorOptics:
I see. Any different setups would have to set their own command. Like mine, because i have multiple devices with amixer, helper.async grabs "mute" from the last device. This poses a problem because that device doesn't mute anything.

copycat-killer:
You're right, it's a bit inflexible like this.

I readded `togglechannel` back in 156515e.

Issue #300, opened 2017-01-31
Calendar: current date is not being highlighted.


sprnza:
Hi there!
I can't remember exactly the update I've started facing this issue but calendar's stopped highlighting the current date.
I use this code to attach the widget:

```
cal = lain.widgets.calendar({attach_to = {mytextclock}})
```

In a terminal `cal` returns the date highlighted properly.

Just tried a default awesom's config- it behaves the same way- no current date highlighting.

copycat-killer:
Try [this](https://github.com/copycat-killer/lain/commit/32eddd514f8ccc35eeaa128a441f6a47708e6a04).

sprnza:
It's already there. I've updated today morning.

dummys:
Why closing the issue ? It's still not working right ?

sprnza:
@dummys it's another issue was closed. This one is not.

copycat-killer:
> It's already there. I've updated today morning.

Sorry, I mean try reverting to the previous code.

sprnza:
@copycat-killer The previous code doesn't work either

copycat-killer:
Is `/usr/bin/cal --color=always` the right command to have a colorized `cal` input in your distro?

For instance, if you're using Ubuntu, it's `/usr/bin/cal -h` instead.

Or did you do specific `/etc/terminal-colors.d` settings?

sprnza:
I use ArchLinux so `--color=always` is the option I should use. Either `cal` and `cal --color=always` returns the current date highlighted properly in terminal but nothing works properly in the calendar's notification.

copycat-killer:
I'm on Arch too. This is what I use:

```lua
lain.widgets.calendar {
    cal = "/usr/bin/cal --color=always",
    attach_to = { mytextclock }
}
```

and it works (simple `cal` doesn't).

You can try adjusting the matching pattern in Lua:

```lua
$ lua
> cal = io.popen("/usr/bin/cal --color=always"):read("*a")
> cal:gsub("%c%[%d+[m]?%d+%c%[%d+[m]?", "DID I PICK IT?")
```

If you read the blurb, then it's working.

Of course, I'll work on a solution myself, but in the meantime you can test.

sprnza:
It's working in lua, but not in Awesome, the calendar doesn't have the highlighting. I just tried a default config(with default theme) with your snippet added. It's not working.

dummys:
Not working in my setup too.
Arch linux as well.

[...]

PR #303, opened 2017-02-01
Fix calender widget not highlighting date #300

2009:
- os.date was returning the day as 01 which meant the calendar icon was looking for 01.png instead of 1.png
- looks like the cal command pads the singe didgets which the gsub didn't account for


copycat-killer:
I was already on this, but thanks.

This is a change in the Lua `os` library.

PR #307, opened 2017-02-02
pulseaudio widget improvements

Raumzeitkeks:
(Preamble: This is my first pull request, so please be forgiving if I do some mistakes. -.-)

Hi there,
I wanted to use two pulseaudio widgets simultaneously in order to monitor the microphone sensivity as well, but the implementation did not handle multiple pulseaudio instances correctly. So I fixed this and also added an easy method to switch between the default pulseaudio sink (speaker) and source (microphone). From there on I also did some more improvements.

All in all I changed the following:
- Correct handling of multiple widget instances.
- Added option for source monitoring.
- Documented the sed command, as this will greatly aid anybody not familiar with sed. (like e.g. myself until just then)
- Created a default setting.
- Added tooltip support and a default tooltip.

Tested it myself using 
```
awesome v4.0 (Harder, Better, Faster, Stronger)
 • Compiled against Lua 5.3.3 (running with Lua 5.3)
 • D-Bus support: ✔
 • execinfo support: ✔
 • RandR 1.5 support: ✔
 • LGI version: 0.9.1
```

### DIFF
--- a/widgets/pulseaudio.lua
+++ b/widgets/pulseaudio.lua
@@ -16,9 +16,10 @@ local setmetatable = setmetatable
 
 -- PulseAudio volume
 -- lain.widgets.pulseaudio
-local pulseaudio = {}
+
 
 local function worker(args)
+   local pulseaudio = {}
    local args        = args or {}
    local timeout     = args.timeout or 5
    local settings    = args.settings or function() end
@@ -58,4 +59,4 @@ local function worker(args)
    return pulseaudio
 end
 
-return setmetatable(pulseaudio, { __call = function(_, ...) return worker(...) end })
+return setmetatable({}, { __call = function(_, ...) return worker(...) end })
--- a/widgets/pulseaudio.lua
+++ b/widgets/pulseaudio.lua
@@ -15,16 +15,16 @@ local string       = { gmatch = string.gmatch,
 -- PulseAudio volume
 -- lain.widgets.pulseaudio
 
-
 local function factory(args)
-   local pulseaudio = {}
-   local args        = args or {}
-   local devicetype  = args.devicetype or "sink"
-   local timeout     = args.timeout or 5
-   local settings    = args.settings or function() end
-   local scallback   = args.scallback
+    local pulseaudio = {}
+    
+    local args        = args or {}
+    local devicetype  = args.devicetype or "sink"
+    local timeout     = args.timeout or 5
+    local settings    = args.settings or function() end
+    local scallback   = args.scallback
 
-   pulseaudio.cmd    = args.cmd or ("pacmd list-" .. devicetype .. "s | sed -n -e '/* index:/,/index:/ !d' -e '/* index:/ p' -e '/base volume:/ d' -e '/volume:/ p' -e '/muted:/ p' -e '/device\\.string/ p'")
+    pulseaudio.cmd    = args.cmd or ("pacmd list-" .. devicetype .. "s | sed -n -e '/* index:/,/index:/ !d' -e '/* index:/ p' -e '/base volume:/ d' -e '/volume:/ p' -e '/muted:/ p' -e '/device\\.string/ p'")
     -- sed script explanation:
     --  '/* index:/,/index:/ !d' removes all lines outside of the range between the line containing '* index:' and the next line containing 'index:'. '* index:' denotes the beginning of the default device section.
     --  '/* index:/ p' prints the line containing '* index:'.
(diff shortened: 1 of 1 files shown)


widgets/pulseaudio.lua
```diff
@@ -10,15 +10,13 @@ local helpers      = require("lain.helpers")
...
-                       match  = string.match,
-                       format = string.format }
-local setmetatable = setmetatable
+                       match  = string.match }
 
 -- PulseAudio volume
 -- lain.widgets.pulseaudio
 
 
-local function worker(args)
```
copycat-killer: Either revert to `worker`, or set `factory` for every other widget too.

widgets/pulseaudio.lua
```diff
@@ -24,7 +24,14 @@ local function factory(args)
    local settings    = args.settings or function() end
    local scallback   = args.scallback
 
-   pulseaudio.cmd    = args.cmd or "pacmd list-" .. devicetype .. "s | sed -n -e '0,/*/d' -e '/base volume/d' -e '/volume:/p' -e '/muted:/p' -e '/device\\.string/p'"
+   pulseaudio.cmd    = args.cmd or ("pacmd list-" .. devicetype .. "s | sed -n -e '/* index:/,/index:/ !d' -e '/* index:/ p' -e '/base volume:/ d' -e '/volume:/ p' -e '/muted:/ p' -e '/device\\.string/ p'")
```
copycat-killer: The command is long, but is very simple too.

If you want to document it, put the explanation in the wiki instead.

widgets/pulseaudio.lua
```diff
@@ -31,7 +31,15 @@ local function factory(args)
     --  '/volume:/ p' prints the line containing 'volume:'.
     --  '/muted:/ p' prints the line containing 'muted:'.
     --  '/device\\.string/ p' prints the line containing 'device.string:'.
-
+    
+    local settings    = args.settings or function()
```
copycat-killer: Like `worker`: either add the same functionality to all widgets, or remove it.

I don't agree on this anyway: we can't assume a default settings, nor is that difficoult to write a `settings` function.

widgets/pulseaudio.lua
```diff
@@ -33,44 +33,49 @@ local function factory(args)
     --  '/device\\.string/ p' prints the line containing 'device.string:'.
     
     local settings    = args.settings or function()
+            widgettext  = volume_now.left .. "%"
```
copycat-killer: This too is just a personal settings.

You can just add the tooltip to `pulseaudio.widget` from `rc.lua`.

Anyway, you can write it in the wiki entry, in a dedicated section.

widgets/pulseaudio.lua
Raumzeitkeks: Will revert this to `worker` due to a lack of time at the moment.
But I'll try to create another pull request fixing the multiple instances 'bug' in all widgets as soon as I get to it.

widgets/pulseaudio.lua
Raumzeitkeks: Alright, will move the comment to the wiki.

How about the change from `-e '0,/*/d'` to `-e '/* index:/,/index:/ !d' -e '/* index:/ p'`? Are you fine with that? I found it a bit clearer to understand and this will also strip all device sections after the default one. But I'm not that familiar with sed and don't know if this is robust.

widgets/pulseaudio.lua
```diff
@@ -33,44 +33,49 @@ local function factory(args)
...
+                tooltiptext = "Mute/" .. tooltiptext
             end
+            widget:set_text(widgettext)
+            tooltip:set_text(tooltiptext)
         end
-        
+    
     pulseaudio.widget = wibox.widget.textbox()
-
+    pulseaudio.tooltip = awful.tooltip({ objects = { pulseaudio.widget } })
```
copycat-killer: This is again just a personal settings: you can add it to `pulseaudio.widget` from `rc.lua`.

And you can write this too in the wiki entry, in a dedicated section.

widgets/pulseaudio.lua
Raumzeitkeks: But unless most other widgets this won't show _anything_ unless the settings function is defined.

copycat-killer: The `settings` function is intended to be always defined, if you use the `.widget`.

widgets/pulseaudio.lua
copycat-killer: If it's semantically equivalent, then it has the same robustness.

widgets/pulseaudio.lua
Raumzeitkeks: Agreed with you in the first place, but on a second thought defining the tooltip here really makes sense though. Otherwise it becomes impossible to _easily_ update it's contents in the settings function, I guess. Also it doesn't actually hurt performance, does it?

copycat-killer: I don't understand how should the tooltip help you in defining the widget settings.

On the performance side, you're still adding a new object and using `awful`, so even if minimal, it's there.

Raumzeitkeks: I meant that it would be difficult to define the tooltip in the `rc.lua` and then keep it's contents synced to the widget. On the contrary by defining it right here it is easily accessible from the `settings` function.

copycat-killer: Yeah, you're right.

We need a tooltip for `alsa` too then. 😈

copycat-killer: Actually, no.

You can define a tooltip before the widget and make `settings` set it. [Example](https://github.com/copycat-killer/awesome-copycats/blob/master/themes/vertex/theme.lua#L127).

copycat-killer:
Approved commits have been merged in a274576.

PR #310, opened 2017-02-06
Weather widget broken again

trap000d:
I've found you've removed all calls of ```read_pipe``` from code. Unfortunately it's caused a wrong time calculation in weather widget. So here is proposed fix 
1) a typo: one should subtract offset from local midnight to calculate UTC midnight
2) lua ```os.time``` behavior differs from system utility ```date```
``` date -u -d 'today 00:00:00' +'%%s' ``` calculates unix seconds for midnight in UTC time zone location, whereas os.time does it for midnight in local time zone. At some conditions (in my place - every day after 1PM) offset between these times becomes more then 24 hours and ````utc_m``` begins to point to yesterday.

### DIFF
--- a/widgets/weather.lua
+++ b/widgets/weather.lua
@@ -130,7 +130,13 @@ local function worker(args)
                 local icon    = weather_now["weather"][1]["icon"]
                 local loc_m   = os.time { year = os.date("%Y"), month = os.date("%m"), day = os.date("%d"), hour = 0 }
                 local offset  = utc_offset()
-                local utc_m   = loc_m + offset
+                local utc_m   = loc_m - offset
+
+                if offset > 0 and (now - utc_m)>=86400 then
+                    utc_m = utc_m + 86400
+                elseif offset < 0 and (utc_m - now)>=86400 then
+                    utc_m = utc_m - 86400
+                end
 
                 -- if we are 1 day after the GMT, return 1 day back, and viceversa
                 if offset > 0 and loc_m >= utc_m then


copycat-killer:
This PR is a mess. Can you please rebase it, so I can understand the changes? Thanks.

trap000d:
Hi Luke,

Sorry for the mess. Though I've rebased forked copy of original repository, git still shows all history of my commits. I suspect the shortest way to "resolve" this is re-forking from scratch. (https://xkcd.com/1597/). However I'll make another one or two attempts to get rid of that.

The commit  with actual fix is here: https://github.com/copycat-killer/lain/pull/310/commits/a0da152bb39c5c3ae239a5743c40cb35038c4a21

Regards,

copycat-killer:
Thanks.

copycat-killer:
@trap000d I changed the logic: instead of doing all the work, I defined `now` time in UTC just like OpenWeatherMap's. Can you please test a5b05fa? Thank you.

trap000d:
Hi Luke,
It doesn't show a proper icon after this update. Either you have to roll it back, or give me a several days (perhaps a week) to make it work with UTC (I'm a bit busy now, so be able to play with it on weekend only)

copycat-killer:
I can roll back, but then I'd experience the icon issue. We have to find a way through.

copycat-killer:
Rolled back, I'll try to solve my issue without altering the current logic.

trap000d:
Hi Luke,
I've created a pull request with fix: https://github.com/copycat-killer/lain/pull/360/commits/42f103141854afe104a2a4cd2f22233f50f3cdcd
Looks like utc_offset is obsolete now, so you might remove it from code

trap000d:
Unfortunately it's not possible to work with UTC only: OWM returns all in UTC however you need to represent it to user in particular location, so you have to convert it to local time. Though conversion looks trivial (utc+offset), in some conditions difference between UTC time from OWM and local time might be more then 24 hours, and then.... https://xkcd.com/1883/

PR #313, opened 2017-02-08
Set widget item in pulseaudio table

flub:
Without this the line below which exposes the widget variable to
the settings function as pulseaudio.widget will end up setting
the widget to nil, which is not very helpful.


flub:
Oops, seems I was not working from the most up-to-date version and this was fixed already.  Apologies for the noise.

PR #316, opened 2017-02-17
Support for unix socket in MPD widget

martindube:
Hi,
This is a small suggestion to support Unix Socket in MPD. 
Thank you,

### DIFF
--- a/widget/mpd.lua
+++ b/widget/mpd.lua
@@ -27,6 +27,7 @@ local function factory(args)
     local args          = args or {}
     local timeout       = args.timeout or 2
     local password      = (args.password and #args.password > 0 and string.format("password %s\\n", args.password)) or ""
+    local socket        = args.socket or os.getenv("HOME") .. "/.config/mpd/socket"
     local host          = args.host or "127.0.0.1"
     local port          = args.port or "6600"
     local music_dir     = args.music_dir or os.getenv("HOME") .. "/Music"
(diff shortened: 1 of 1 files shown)


copycat-killer:
This introduces GNU netcat as dependency, which is not part of every distro core installation.

Please, investigate if is there a Pure Lua/Awesome API implementation.

Also, note that if you want to use sockets, [this recipe](https://awesomewm.org/recipes/mpc) might be of interest.

Veratil:
An alternative to netcat I've seen is piping into `/dev/tcp/...`. Something like this:
`echo 'blah blah' > /dev/tcp/address/port`

copycat-killer:
Then implement and test it in your fork, please.

Unode:
[Maybe this?](https://github.com/awesomeWM/awesome-www/blob/master/recipes/mpc.lua)

Unode:
[Maybe this?](https://github.com/awesomeWM/awesome-www/blob/master/recipes/mpc.lua)

copycat-killer:
@Unode Another thing I want to do for `lain` first release is to get rid of `curl` in `mpd`, `imap` and `weather`, in order to remove any external dependency.

Eventually, this PR might be overridden, or pushed, along with the `curl`-dependent code, as a recipe on the awesome site.

Unode:
@copycat-killer by "get rid" I hope you mean replacing them with lua native api's such as [luacurl](https://github.com/Lua-cURL/Lua-cURLv3).
If you remove the widgets people will miss them :)

copycat-killer:
@Unode By "get rid" I'm referring to `curl`.

I want to replace it with GLib via `lua-lgi` (which is an Awesome dependency).

PR #318, opened 2017-02-20
Always display transfer rates as string

Unode:
When all devices idle the widget returns 2 integer values.
When still something is measured the widget behaves in a rather "jumpy" way switching between "0 0" and "0.1 0.0" or similar values which makes the entire widget resize itself all the time.

The affected if clause was trying to avoid a tiny performance impact but the change shouldn't be noticeable.

This patch removes this check and always returns a float as a string.

### DIFF
--- a/widget/net.lua
+++ b/widget/net.lua
@@ -103,12 +103,10 @@ local function factory(args)
             -- totals across all specified devices
         end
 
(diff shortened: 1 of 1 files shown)


Unode:
Thanks

PR #335, opened 2017-03-30
make battery low/crit levels configurable

jandob:
I will also update the wiki and add a note that these percentages are calculated based on all batteries combined.

### DIFF
--- a/widget/bat.lua
+++ b/widget/bat.lua
@@ -1,10 +1,10 @@
 
 --[[
-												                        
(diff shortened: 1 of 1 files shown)


copycat-killer:
Instead of 2 variables, you can define `n_perc`, a table of 2 integers, respectively critical and low percentage notification values. Example:

```lua
local mybattery = lain.widget.bat {
    n_perc = {10, 25},
    -- [...]
}
```

Issue #272, opened 2017-01-22
fs: pure dfs implementation


lcpz:
Remove `scripts/` definitely and switch to a pure Lua implementation.

aajjbb:
By 'pure Lua' implementation you mean "No parsing `df`" ?

copycat-killer:
No parsing anything at all, but using Lua and Awesome APIs.

I was thinking about a wibox to toggle, filled with progress bars.

psychon:
I'm not yet sure how to get a list of mount points, but at least this is how you can get some disk usage:
```lua
gio = require("lgi").Gio
file = gio.File.new_for_path(".")
size, free, used = gio.FILE_ATTRIBUTE_FILESYSTEM_SIZE, gio.FILE_ATTRIBUTE_FILESYSTEM_FREE, gio.FILE_ATTRIBUTE_FILESYSTEM_USED
query = size..","..free..","..used
info = file:query_filesystem_info(query)
print(info:get_attribute_uint64(size), info:get_attribute_uint64(free), info:get_attribute_uint64(used))
```

psychon:
I can only offer the following, but it does not list all mount points, only some (yes, I know that this queries free but does not use it):
```lua
local Gio = require("lgi").Gio
local volume_monitor = Gio.VolumeMonitor.get()
local query_size = Gio.FILE_ATTRIBUTE_FILESYSTEM_SIZE
local query_free = Gio.FILE_ATTRIBUTE_FILESYSTEM_FREE
local query_used = Gio.FILE_ATTRIBUTE_FILESYSTEM_USED
local query = query_size .. "," .. query_free .. "," .. query_used

for _, mount in ipairs(volume_monitor:get_mounts()) do
	local root = mount:get_root()
	local info = root:query_filesystem_info(query)
	local size = info:get_attribute_uint64(query_size)
	local free = info:get_attribute_uint64(query_free)
	local used = info:get_attribute_uint64(query_used)
	local percentage = 100 * used / size
	print(string.format("%s '%s' has %.3f%% used", root:get_path(), mount:get_name(), percentage))
end
```
Output:
```
/ 'Filesystem root' has 21.000% used
/boot/efi 'efi' has 0.025% used
/mnt/space 'space' has 7.938% used
```

Unode:
Do we have any progress on this?

Currently I'm experiencing problems with the notification pop-up.
Running the `dfs` script manually works fine but the pop-up only shows:

                                                           Used     Free    Total

[...]

PR #338, opened 2017-04-27
scripts/dfs: specify bash instead of sh

klaernie:
The `sh` on e.g. Debian is `dash`, but the script requires bash features by using `+=`. Until #272 is complete I think this is the best way to ensure that dfs works by default (excluding the gawk requirement).


copycat-killer:
I will remove this script ASAP.

PR #339, opened 2017-05-04
quake: add a new option `onlyone`.

exzombie:
This option allows using a single instance of the app with multiple
screens instead of having a separate app for each screen. It works
better if there is only one instance of the quake object, i.e.

    local quake = lain.util.quake({onlyone=true})

and

    awful.key({ modkey, }, "z", function () quake:toggle() end),

### DIFF
--- a/util/quake.lua
+++ b/util/quake.lua
@@ -27,6 +27,8 @@ local quake = {}
 
 function quake:display()
     if self.followtag then self.screen = awful.screen.focused() end
+    local toscan = self.screen
+    if self.onlyone then toscan = nil end
 
     -- First, we locate the client
     local client = nil
@@ -34,7 +36,7 @@ function quake:display()
     for c in awful.client.iterate(function (c)
         -- c.name may be changed!
         return c.instance == self.name
-    end, nil, self.screen)
+    end, nil, toscan)
     do
         i = i + 1
         if i == 1 then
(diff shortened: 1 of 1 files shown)


copycat-killer:
@liquiddandruff reported [here](https://github.com/copycat-killer/awesome-copycats/issues/132#issuecomment-327130937) that the option seems not working with his setup.

Since I'm not a multiscreen user, and you introduced the PR, I'd like to have your help, @exzombie.

---

@liquiddandruff when you spawn quake on the other screen, is the previous instance visible?

exzombie:
@liquiddandruff, do you construct your `quake` object inside `connect_for_each_screen` or globally? Option `onlyone` only makes sense when having a global `quake` object. If you make one for each screen, it can have no effect. In fact, I intended to update the wiki when this PR was merged, but forgot.

@copycat-killer, I thought that the default behaviour of spawning a new terminal for each screen was a bit strange, now I know why :) . I decided to introduce a new parameter in order not to change behaviour for existing configurations. But since using `onlyone` also requires moving `quake` to global context, would it make sense to remove the `onlyone` parameter and just always scan all screens? One can then choose whether to have a global or per-screen `quake` object, a simple documentation change.

copycat-killer:
You're right, my head is elsewhere these days.

I'll remove the option and change documentation.

liquiddandruff:
When I spawn quake on the other screen, the previous instance is still visible, yes. And I did construct the quake object for each screen instead of globally. It seems the problem has been identified; am eager for the next revision, and thanks to you both!

exzombie:
I took the liberty of updating the wiki as keybinding has to be done differently. I also removed `onlyone` from the table. But you still need to fix the code: on line 36, you iterate over `self.screen`; use `nil` instead. I'm glad this will be solved without a superfluous parameter :) .

PR #340, opened 2017-05-06
widget/bat: Use BAT1 instead of BAT0

Sudokamikaze:
Some laptops use BAT1 instead BAT0

Or create another if exception If there  BAT1

### DIFF
--- a/widget/bat.lua
+++ b/widget/bat.lua
@@ -1,10 +1,10 @@
 
 --[[
-												                        
-	 Licensed under GNU General Public License v2 
-	  * (c) 2013,      Luke Bonham                
-	  * (c) 2010-2012, Peter Hofmann              
-												                        
+
+	 Licensed under GNU General Public License v2
+	  * (c) 2013,      Luke Bonham
+	  * (c) 2010-2012, Peter Hofmann
+
 --]]
 
 local first_line = require("lain.helpers").first_line
@@ -26,7 +26,7 @@ local function factory(args)
     local bat       = { widget = wibox.widget.textbox() }
     local args      = args or {}
     local timeout   = args.timeout or 30
-    local batteries = args.batteries or (args.battery and {args.battery}) or {"BAT0"}
+    local batteries = args.batteries or (args.battery and {args.battery}) or {"BAT1"}
     local ac        = args.ac or "AC0"
     local notify    = args.notify or "on"
     local n_perc    = args.n_perc or { 5, 15 }


aajjbb:
But it would break lain for those who use it with BAT0. You can explicity change your `bat`in the creation of the widget. 

One better way to solve it would be a way to discovery the battery id automatically.

Sudokamikaze:
@aajjbb 
I think, better to discover automatically, in example like `BATVAR=$(ls /sys/class/power_supply | grep BAT)` in bash

aajjbb:
It sounds good, the only issue is how would we call this `ls`, there's a lot of criticism in awesome on calling things like that in a sincronous (with `io.popen`), I don't see harm in it for simples taks like that but, better wait for what @copycat-killer thinks about changing it.

Sudokamikaze:
@aajjbb Maybe use `os.execute(string.format("" ))` but how to call BARVAR variable?

aajjbb:
The issue is that `os.execute` is synchronous as well, that's my concern. If being synchronous isn't a concern, using `io.popen('ls /sys/class/power_supply | grep BAT')` does the job here.

copycat-killer:
I can't merge this.

I can however introduce a function like [`net` one](https://github.com/copycat-killer/lain/blob/master/widget/net.lua#L31-L36) to "autodetect" the battery ID, which is asynchronous.

The idea is to get a table of strings from `BAT*` directories in `/sys/class/power_supply`, so that this will work with any number of batteries available, while the first `AC*` directory found will be `ac` variable.

I'll leave this open until it's done.

Sudokamikaze:
@copycat-killer Ok, i'll wait for your commits, thanks for attention

fleekix:
Will this auto detect multiple batteries too? (crosses fingers haha)*cough thinkpad x250*

copycat-killer:
@fleekix It should: do you have your battery dirs all listed in `/sys/class/power_supply`?

PS: Check also [this](https://github.com/copycat-killer/lain/wiki/tpbat).

fleekix:
Hey thanks Luke. 
I have both batteries listed in that dir.

I should stop being lazy and just get it figured out, I had it done at one point but did a fresh install and since then I haven't done it (I have forgotten what I did to show both batteries) 

I will look at the thinkpad specific module. Right now I have been trying to figure out why my drop down term is white and my normal term (xterm) is dark. 

Here's to tinkering!
Cheers!

lcpz:
Why did you close this?

I know, it has been a lot of time, but it's still part of my plan for the first stable release.

Sudokamikaze:
@lcpz I thought it was empty pull request, sorry about this

nonchip:
why call `ls` anyway and not just use luafilesystem or `gears.filesystem` to list the directory? way less overhead.

lcpz:
> way less overhead

You're right. I've started recently to refactor some code (see `widget.fs` for instance) in order to use Awesome or Gio facilities. Can you please do a general review and tell me if you spot any other parts in this repository which can improved likewise?

In the meantime, I'll open an issue form for this.

nonchip:
reviewing overhead wise, except from trying to minimize spawning processes, the most improvements could be got imho by modifying `helpers.first_line` to run `io.lines` just once instead of reading the whole file into a lines-table and then pulling the first element.
couldn't find anything else yet, but if I find anything else in the future I'll file an issue and/or PR :)

lcpz:
> except from trying to minimize spawning processes

What are you referring to?

lcpz:
I finally introduced battery autodetection. Feel free to test it and give me some feedback.

PR #341, opened 2017-05-16
Short circuit notification check if fs_now.used is not a number

olidee:
I was getting a "tried to compare number with nil" error, presumably because fs_now.used was "N/A".


copycat-killer:
`tonumber("N/A")` is `nil`, so your current patch still does the same comparison.

Remove the `> 0`.

PR #342, opened 2017-05-24
Fix GPMDP widget

simbuerg:
The GPMDP widget fails to do anything because of 2 issues:
    1) The scope of gpm_now is broken.
    2) The format string in the curl call expects a number, but an URL is not a number.

### DIFF
--- a/widget/contrib/gpmdp.lua
+++ b/widget/contrib/gpmdp.lua
@@ -67,7 +67,7 @@ local function factory(args)
 
                 if followtag then gpmdp_notification_preset.screen = focused() end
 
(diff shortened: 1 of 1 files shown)


copycat-killer:
I want to turn this into a [watcher](https://awesomewm.org/recipes/watch) for various reasons, but I don't have the time right now. Could you do it? It should be not too much difficoult.

simbuerg:
That assumes that I know a single bit about lua (haha) :-) I can give it a try.

copycat-killer:
Thanks, I can give you advices if needed. Meanwhile, I'll keep these issues open until we solve the question.

copycat-killer:
https://github.com/copycat-killer/lain/pull/354#issuecomment-322164953

PR #349, opened 2017-07-14
mpd: Allow mpd connection via UNIX socket

FloatingGhost:
Previously, the mpd widget couldn't communicate via a UNIX socket, leading to it being basically useless in that case.

We can use exactly the same command sent through `socat` to fix this, which this implements.

### DIFF
--- a/helpers.lua
+++ b/helpers.lua
@@ -15,7 +15,7 @@ local io         = { lines = io.lines,
 local rawget     = rawget
 local table      = { sort  = table.sort }
 
-local gio = require("lgi").Gio
+local Gio = require("lgi").Gio
 
 -- Lain helper functions for internal use
 -- lain.helpers
@@ -153,54 +153,42 @@ function generate_buffer(buffer_length)
     return recv_buffer
 end
 
--- Send some data to a UNIX socket
+-- Wrapper function to send to any sort of address
+function helpers.send_to_address(host, port, data, buffer_length, callback)
+    -- Have a default buffer length of 1000
+    if not buffer_length then
+      buffer_length = 1000
+    end
 
-function helpers.send_to_unix_socket(unix_path, data, buffer_length) 
-    -- Create a buffer to recieve into
+    -- Generate a buffer to store our result in
     local recv_buffer = generate_buffer(buffer_length)
 
-    -- Create a socket to send some data from
-    local sock = gio.Socket.new(gio.SocketFamily.UNIX,
-                                gio.SocketType.STREAM,
-                                gio.SocketProtocol.DEFAULT)
+    -- Check if we should be sending to a socket or an IP
+    local is_socket = (string.sub(host, 1, 1) == "/")
 
-    -- Create a socket address to connect to
-    local addr = gio.UnixSocketAddress.new(unix_path)
-    -- Connect across
-    sock:connect(addr)
-    -- Send our message
-    sock:send(data)
-    -- Pull back anything it sends back
-    sock:receive(recv_buffer)
-    
-    sock:close()
+    -- Create a client to listen and send with
+    local client = Gio.SocketClient()
 
-    return recv_buffer    
+    local addr
+
+    if is_socket then
+        addr = Gio.UnixSocketAddress.new(host)
+    else
+        local inet_addr = gio.InetAddress.new_from_string(host)
+        addr = Gio.InetSocketAddress.new(inet_addr, port)
+    end
+
+    local conn = client:connect(addr)
+
+    local input_stream = conn:get_output_stream()
+    local output_stream = conn:get_input_stream()
+
+    input_stream:write(data)
+    output_stream:read(recv_buffer)
+    output_stream:read(recv_buffer)
+
+    callback(recv_buffer)
 end
-
--- Send some data to an IPv4 socket
-function helpers.send_to_ip_address(ip_address, port, data, buffer_length)
-    -- Create a buffer to recieve into
-    local recv_buffer = generate_buffer(buffer_length)
-    
-    -- Create a socket to send some data from
-    local sock = gio.Socket.new(gio.SocketFamily.IPV4,
-                                gio.SocketType.STREAM,
-                                gio.SocketProtocol.DEFAULT)
-
-    -- Create a socket address to connect to
-    local inet_addr = gio.InetAddress.new_from_string(ip_address)
-    local addr = gio.InetSocketAddress.new(inet_addr, port)
-    -- Connect across
-    sock:connect(addr)
-    -- Send our message
-    sock:send(data)
-    -- Pull back anything it sends back
-    sock:receive(recv_buffer)
-    sock:close()
-    return recv_buffer
-end
-
 -- }}} 
 
 -- {{{ Misc
--- a/helpers.lua
+++ b/helpers.lua
@@ -144,10 +144,10 @@ end
 -- Probably just gonna leave it local for now
 function generate_buffer(buffer_length)
     -- First, create an output buffer to recieve from
-    local recv_buffer = ""
+    local recv_buffer = " "
     -- We'll need to allocate `buffer_length` bytes to it
     for i=1,buffer_length do
-        local recv_buffer = recv_buffer .. " "
+        recv_buffer = recv_buffer .. " "
     end
 
     return recv_buffer
@@ -189,9 +189,8 @@ function helpers.send_to_ip_address(ip_address, port, data, buffer_length)
                                 gio.SocketProtocol.DEFAULT)
 
     -- Create a socket address to connect to
-    local inet_addr = Gio.InetAddress.new_from_string(ip_address)
-    local addr = gio.UnixSocketAddress.new(inet_addr, port)
-
+    local inet_addr = gio.InetAddress.new_from_string(ip_address)
+    local addr = gio.InetSocketAddress.new(inet_addr, port)
     -- Connect across
     sock:connect(addr)
 
--- a/helpers.lua
+++ b/helpers.lua
@@ -193,13 +193,11 @@ function helpers.send_to_ip_address(ip_address, port, data, buffer_length)
     local addr = gio.InetSocketAddress.new(inet_addr, port)
     -- Connect across
     sock:connect(addr)
-
     -- Send our message
     sock:send(data)
-
     -- Pull back anything it sends back
     sock:receive(recv_buffer)
-
+    sock:close()
     return recv_buffer
 end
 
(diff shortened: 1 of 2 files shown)


copycat-killer:
Thanks for your time, but this has the same points and problems of #316, since `socat` is a dependency.

[This](https://github.com/copycat-killer/lain/pull/316#issuecomment-282473275) is my desideratum. In short, I want to remove `curl` as dependency and use sockets for network-based widgets, but I don't have the time at the moment. I'll keep this open too until it's done.

FloatingGhost:
ok, I've added a basic implementation of a send to unix socket w/ lua-lgi (because I had no internet and had nothing else to do ;0)

The main issue is going to be maintaining that async format - I'm really rather new to lua so idk if there's an easy way to do it

helpers.lua
```diff
@@ -15,6 +15,8 @@ local io         = { lines = io.lines,
 local rawget     = rawget
 local table      = { sort  = table.sort }
 
+local gio = require("lgi").Gio
```
copycat-killer: Please, rename the variable `Gio` for consistency.

copycat-killer:
I appreciate your efforts, but don't reinvent the wheel. 

My aim is generalise [this](https://github.com/awesomeWM/awesome-www/blob/master/recipes/mpc.lua) as a `lain.helper` (indeed, you good the right idea), but using [`Gio.UnixInputStream`](https://lazka.github.io/pgi-docs/Gio-2.0/classes/UnixInputStream.html) instead of `Gio.DataInputStream`.

FloatingGhost:
Literally just as I get it to work properly over Gio streams as well >__<

Eh I'll see if I can get that to work instead, SocketClients make me want to cry

helpers.lua
```diff
@@ -134,6 +136,61 @@ end
...
+function generate_buffer(buffer_length)
+    -- First, create an output buffer to recieve from
+    local recv_buffer = " "
+    -- We'll need to allocate `buffer_length` bytes to it
+    for i=1,buffer_length do
+        recv_buffer = recv_buffer .. " "
+    end
+
+    return recv_buffer
+end
```
psychon: `return string.rep(" ", buffer_length + 1)` (the `+1` is there to match the behaviour of your code, but I don't know if your code really should do that)

helpers.lua
```diff
@@ -134,6 +136,61 @@ end
...
+    local conn = client:connect(addr)
+
+    local input_stream = conn:get_output_stream()
+    local output_stream = conn:get_input_stream()
+
+    input_stream:write(data)
+    output_stream:read(recv_buffer)
+    output_stream:read(recv_buffer)
+
+    callback(recv_buffer)
```
psychon: Hm... this will actually "break Lua". Lua assumes that strings are immutable. So, e.g., if any other piece of code constructed the same number of blanks before, the received data will also end up in there.

Also, you should `conn:close()` when done.

(Note that I just looked over this in a drive-by fashion, @copycat-killer gets to decide things and I have no say in anything here)

FloatingGhost:
heh, it must be pretty obvious that I'd never used lua before writing this I guess

T'is a silly language
I'm using the PR version on my system and it doesn't seem to... break, so I guess there's that

helpers.lua
copycat-killer: @psychon I appreciate any review of yours, for me you are free to come by as you like.

razamatan:
what more needs to be done here?  just a rebase w/ current head to get rid of the merge conflicts?

lcpz:
@razamatan See https://github.com/lcpz/lain/pull/349#issuecomment-315391410

psychon:
@lcpz Do you know about https://awesomewm.org/recipes/mpc.lua already? It connects to mpd over either TCP or a unix socket and provides an asynchronous API for talking to the daemon.

lcpz:
@psychon The idea was indeed to readapt that. No need to reinvent the wheel.

razamatan:
ok.. i'm taking this up.  do you want a stateful socket that tries to reuse the connection as long as possible, or do you want to go the direction that @FloatingGhost is headed with reconnecting every time for a request/response?

lcpz:
@razamatan The idea is still [this](https://github.com/lcpz/lain/pull/349#issuecomment-531862205). Hence, sockets. Try to work with the above-mentioned code, it should be easy.

razamatan:
stateful sockets, got it.. i started this and i'm working it out now.

razamatan:
hmm... while making the socket helper/class generalizable, the socket protocol is too ambiguous to make the helper/class useful beyond some initial connection boilerplate.  the central read loop needs to be implemented per protocol/usage: http things would go one way, mpd has it's own client-initiated conversational api, etc. etc.  once i gutted the main read loop to be a user-defined callback, it really isn't much.  i could make it a base class for each client/protocol to implement/inherit, but i'm not sure that's what you really wanted here.

also, why the preference for UnixInputStream over DataInputStream?

lcpz:
> i could make it a base class for each client/protocol to implement/inherit, but i'm not sure that's what you really wanted here

You could create a protocol submodule which, for the time being, provides MPD connections only.

In the future, we can also add HTTP(S) and IMAP (and finally remove Curl as a dependency).

Feel free to push WIP commits, if you want my commentary as you work.

We can squash them when you are done.

> also, why the preference for UnixInputStream over DataInputStream?

Because the former is [pollable](https://lazka.github.io/pgi-docs/Gio-2.0/classes/UnixInputStream.html), while the latter is [not](https://lazka.github.io/pgi-docs/Gio-2.0/classes/DataInputStream.html). But it is no religion: feel free to show me that DataInputStreams suit better.

razamatan:
qq.. what's more important, not taking on any dependency at all, or not adding a non-lua approach?  the more and more i'm dealing w/ this lgi wrapped GIO stuff, the more i am inclined to say that luasocket may be a better fit here.

psychon:
@razamatan Anything I can help you with in LGI? Does perhaps already https://github.com/pavouk/lgi/pull/243 help you?

Another helpful pointer could be `awful.spawn.read_lines`: https://github.com/awesomeWM/awesome/blob/87928befe2dc981bfdbca59c0474c0ff5d39544f/lib/awful/spawn.lua#L504
This gets an input stream and read lines from it. Each line is passed to a callback function. Another callback is called on EOF.

razamatan:
i have things working fine for mpd using lgi and UnixInputStream as requested, but for http, it's a bit more complicated.  in order to be robust against all the different possible protocol versions, ssl, redirects, and all the features of the http spec, it's redundant to roll yet another http client.  being open source, i could just grab their http.lua and just refactor it to be only client focused and gio based.  still, more time taken.

the PITA for lgi is that it doesn't have any method level documentation (showing input and return values) since it does things to make the lua binding more lua like.  the overview and guide explain things, but it's much, much easier if there would be just something that shows all the available methods and their signatures.  i've been relying on pgi docs online and testing things to verify behaviors.

psychon:
I think "the way to go" with Gio and http is to just open an URL as a file, but that requires a new package whose name I forgot (something like gio-networking). With that, https://developer.gnome.org/gio/stable/GFile.html#g-file-new-for-uri should just work.

razamatan:
the file abstraction may be fine for terminal rpcs like http requests/responses, but for persistent client/server connections like mpd which prefers keeping the connection alive using idle commands, the file abstraction might not be the best thing.

razamatan:
so work got a bit lighter, so i'm picking this up again.

Gio.File seems to do what we want for most cases (as mentioned by @psychon) that just wants to download a full file, but it requires making sure folks have gvfs installed so gobject-introspection picks up uri support in gio.  the alternative of going w/ the UnixInputStream approach is that all protocols need to be implemented at the socket layer, which is a tax on ongoing maintenance (i was doing this route, but shelved it after psychon's suggestion).  However!  i found that lgi supports libsoup, which is the level of support we care for anyway, so i got that working.

@lcpz are you ok with taking a dep on lgi w/ libsoup?  people who aren't gnome heavy won't have it installed like me (i run a pretty lean X in gentoo).

https://github.com/lcpz/lain/pull/549

lcpz:
@razamatan Thank you for your work, and for sticking around for so long! I appreciate it.

I understand that having no dependencies requires utilities that need to be maintained, which is not reasonable.

Thus, I think `(lib)curl` is a better choice than `libsoup`, as it is more mature, lighter, and [more adopted](https://curl.se/libcurl/libsoup.html).

What do you guys think?

razamatan:
wouldn't using libcurl end up with another dependency on top of it to handle the c <=> lua handling?  i understood https://github.com/lcpz/lain/pull/349#issuecomment-618393393 to limit the number of things that users would need to install on top of awesomewm.  so, lgi => gi => gio|soup was what i thought was wanted.

all this said, unless gnome is thinking of removing libsoup, i don't think it's a bad dependency to take.

lcpz:
I mean to say that there is already `curl` as a dependency, which seems better than `libsoup`, so it's not worth to switch.

But I am open to thoughts.

razamatan:
but the usage of curl as a spawn shell command is the root of the issue, right?  if we want to move to an all lua interface for these rpcs, we would take on luacurl or some other lua wrapper for libcurl as another dependency.

so either some lua->curl wrapper, or making sure folks have libsoup installed so lgi supports it.  my feeling is that people who are already in the world of gnome/gobjects likely will have libsoup installed, but most folks won't have a lua wrapper for libcurl installed.
________________________________

razamatan:
you know what.. i'm going to see if i can just use basic lua -> c behaviors to avoid using a canned lua->libcurl wrapper as a dependency.  i don't think we need a whole lot of functionality.  this way, no new deps, and it'll be easier to migrate from `curl` to libcurl that it provides.

razamatan:
(to clarify the previous comment)

i came to this conclusion b/c https://github.com/Lua-cURL/Lua-cURLv3 (lua-curl; note the hyphen) is the best that's out there for lua and libcurl afaict, and it hasn't been touched in over a year and a half.  luacurl's (non-hyphenated) public repo seems to be no more.

if i don't get something going along the lines suggested in the previous comment that seems easy to maintain in a few more hours, i'll give up and settle on libsoup b/c: 1. gnome supports it 2. gi supports it 3. lgi supports it 4. it works in #549.  this feels much more stabler than relying on some lua-?curl impl...

razamatan:
ok.. i learned how to call c from lua, and i don't want to maintain another lua*curl wrapper.  we don't need another in the world as lua-curl(v3) should be sufficient.  however, i don't want to go off and help maintain that library either.

the lightest maintenance cost is going to be whatever we can get cheaply from lgi, so libsoup as i implemented it is the simplest.  let's go with it.

PR #351, opened 2017-07-20
ability to disable calendar icon

gregflynn:
Adds a new option to the calendar widget to remove the icon from the naughty notification. I was working around this by setting `icons = ' '` but that spits errors into `~/.xsession-errors` every time I view the notification.


copycat-killer:
Now you can set `icons = ''` (empty string) in order to disable it.

gregflynn:
Ah that's more elegant, thanks! Also thanks for updating the wiki too!

PR #362, opened 2017-09-13
Add devicetype to pulsebar to match pulseaudio and update sed command

2009:
It looks like `sink` was only used to display tooltips, but default `cmd` would use the default sink. So if you changed the default the tooltips would be incorrect.

I've updated pulsebar to work similar to the pulseaudio widget with `devicetype` and `device`, and updated the sed cmd which would would never set `device`.


copycat-killer:
I'll remove legacy stuff, since we are git and still pre 1.0.

PR #366, opened 2017-10-05
add center layout that goes left to right

scottgreenup:
I've been using this for a while, an alternative centerwork layout that goes from left to right.

```
+--------------------+---+                                                          
| 1 |                | 6 |                                                          
+---+                +---+                                                          
| 2 |                | 7 |                                                          
+---+  5 and Master  +---+                                                          
| 3 |                | 8 |                                                          
+---+                +---+                                                          
| 4 |                | 9 |                                                          
+---+----------------+---+
```

I duplicated the centerwork.lua file and changed how `do_centerwork` functions to achieve this. I've been using it for a while and hit no snags with it.


layout/centerwork_leftright.lua
```diff
@@ -0,0 +1,137 @@
...
+local naughty = require("naughty")      -- notification library
+
+local function dp(words)
+    naughty.notify({
+        preset = naughty.config.presets.normal,
+        title = "Debug Message",
+        text = words,
+        width = 400
+    })
+end
```
lcpz: Naughty and debugging stuff have to go.

layout/centerwork_leftright.lua
```diff
@@ -0,0 +1,137 @@
...
+
+local floor  = math.floor
+local screen = screen
+
+local centerwork = {
+    name         = "centerwork",
+    horizontal   = { name = "centerworkh" }
+}
```
lcpz: Conflicts with standard centerwork, should have a different name.

lcpz:
Instead of creating a new file, can you modify current `centerwork` to have an optional argument which enables this different placement?

lcpz:
I finally had the occasion to give it a look.

Your alternative seems to behave like the original, at least on my machine.

Moreover, the displacement flow you want is not clear: how many clients should be placed in a row, before switching to the other one? Should this be periodic (X on left then Y on right then repeat) or not (clients after X+Yth one are not arranged)?

What you could do is to *not* duplicate but extend `centerwork` with a different way to arrange rows. Work within this [loop](https://github.com/lcpz/lain/blob/master/layout/centerwork.lua#L70), so that the patch would be independent from orientation (that is, it'd apply to both vertical and horizontal layouts). Open a new PR when you're done.

PR #368, opened 2017-11-18
Show calendar on focused screen

seregaxvm:
Shows calendar on focused screen. Before it was shown on primary display only.


lcpz:
Please set it in [`notification_preset`](https://github.com/lcpz/lain/wiki/calendar#input-table). Thank you.

PR #372, opened 2017-12-13
Wifi indicator using nmcli.

amirulmenjeni:
I couldn't find a wifi indicator widget here so decided to make one with nmcli. Currently the widget shows the SSID and the signal strength to the connected wifi. The user will be able to specify the network device. The user can also set the placeholder for the SSID and signal strength indicator in the case when the  device is not connected.

 Thought?

### DIFF
--- a/widget/contrib/wifi.lua
+++ b/widget/contrib/wifi.lua
@@ -8,8 +8,8 @@ function factory(args)
     local settings  = args.settings or function() end
          
     function wifi.is_connected()
-        conn_flag = io.popen("[[ ! -z `ping -c 1 -w 3 www.duckduckgo.com` ]] && echo true"):read()
-        if conn_flag == "true" then
+        conn_flag = io.popen("[[ ! -z `ping -c 1 -w 3 46.51.218.82` ]] && echo 1"):read()
+        if conn_flag == "1" then
             return true
         end
         return false
@@ -18,18 +18,13 @@ function factory(args)
     function wifi.update()
         wifi_now = {
             ssid = "N/A",
-            signal = "/NA",
-            enable_signal = true,
+            signal = "N/A",
             connected = false
         }
 
         if wifi.is_connected() then
-            wifi_now.ssid = io.popen("nmcli -t -f SSID dev wifi list"):read()
-
-            if wifi_now.enable_signal then
-                wifi_now.signal = io.popen("nmcli -t -f SIGNAL dev wifi list"):read()
-            end
-
+            wifi_now.ssid = io.popen("iwgetid -r"):read()
+            wifi_now.signal = io.popen("nmcli -t -f SIGNAL dev wifi"):read()
             wifi_now.connected = true
         end
 
--- a/widget/contrib/wifi.lua
+++ b/widget/contrib/wifi.lua
@@ -1,5 +1,4 @@
 local awful     = require("awful")
-local naughty   = require("naughty")
 local wibox     = require("wibox")
 local helpers   = require("lain.helpers")
 
--- a/widget/contrib/wifi.lua
+++ b/widget/contrib/wifi.lua
@@ -1,0 +1,45 @@
+local wibox     = require("wibox")
+local helpers   = require("lain.helpers")
+
+function factory(args)
+    local wifi      =  { widget = wibox.widget.textbox() }
+    local args      = args or {}
+    local timeout   = args.timeout or 2
+    local settings  = args.settings or function() end
+         
+    function wifi.is_connected()
+        conn_flag = io.popen("[[ ! -z `ping -c 1 -w 3 www.duckduckgo.com` ]] && echo true"):read()
+        if conn_flag == "true" then
+            return true
+        end
+        return false
+    end
+
+    function wifi.update()
+        wifi_now = {
+            ssid = "N/A",
+            signal = "/NA",
+            enable_signal = true,
+            connected = false
+        }
+
+        if wifi.is_connected() then
+            wifi_now.ssid = io.popen("nmcli -t -f SSID dev wifi list"):read()
+
+            if wifi_now.enable_signal then
+                wifi_now.signal = io.popen("nmcli -t -f SIGNAL dev wifi list"):read()
+            end
+
+            wifi_now.connected = true
+        end
+
+        widget = wifi.widget
+        settings()
+    end
+
+    helpers.newtimer("wifi", timeout, wifi.update)
+
+    return wifi
+end
+
+return factory
--- a/widget/contrib/wifi.lua
+++ b/widget/contrib/wifi.lua
@@ -51,7 +51,7 @@ function factory(args)
 
             awful.spawn.easy_async(wifi.shell_cmd(command.nmcli_wifi),
                 function(stdout, stderr, reason, exit_code)
-                    local ssid, signal = string.match(stdout, ":(.-):(.*)$")
+                    local ssid, signal = string.match(stdout, ":(.-):(%d+)")
                     wifi_now.ssid = ssid
                     wifi_now.signal = signal
                 end
--- a/widget/contrib/wifi.lua
+++ b/widget/contrib/wifi.lua
@@ -32,7 +32,7 @@ function factory(args)
             cmd_sigl = "nmcli -t -f SIGNAL dev wifi list ifname " .. wifi_now.device
             ssid_out = io.popen(cmd_ssid):read()
             signal_out = io.popen(cmd_sigl):read()
-            if not ssid_out == nil and not signal_out == nil then
+            if not (ssid_out == nil or signal_out == nil) then
                 wifi_now.ssid = ssid_out
                 wifi_now.signal = signal_out
             end
--- a/widget/contrib/wifi.lua
+++ b/widget/contrib/wifi.lua
@@ -15,16 +15,27 @@ function factory(args)
         return false
     end
 
+    function wifi.get_device()
+        return io.popen("nmcli -t -f DEVICE dev"):read()
+    end
+
     function wifi.update()
         wifi_now = {
-            ssid = "N/A",
-            signal = "N/A",
+            device = args.device or wifi.get_device(),
+            ssid = args.ssid_placeholder or "N/A",
+            signal = args.signal_placeholder or "N/A",
             connected = false
         }
 
         if wifi.is_connected() then
-            wifi_now.ssid = io.popen("nmcli -t -f SSID dev wifi"):read()
-            wifi_now.signal = io.popen("nmcli -t -f SIGNAL dev wifi"):read()
+            cmd_ssid = "nmcli -t -f SSID dev wifi list ifname " .. wifi_now.device
+            cmd_sigl = "nmcli -t -f SIGNAL dev wifi list ifname " .. wifi_now.device
+            ssid_out = io.popen(cmd_ssid):read()
+            signal_out = io.popen(cmd_sigl):read()
+            if not ssid_out == nil and not signal_out == nil then
+                wifi_now.ssid = ssid_out
+                wifi_now.signal = signal_out
+            end
             wifi_now.connected = true
         end
 
--- a/widget/contrib/wifi.lua
+++ b/widget/contrib/wifi.lua
@@ -23,7 +23,7 @@ function factory(args)
         }
 
         if wifi.is_connected() then
-            wifi_now.ssid = io.popen("iwgetid -r"):read()
+            wifi_now.ssid = io.popen("nmcli -t -f SSID dev wifi"):read()
             wifi_now.signal = io.popen("nmcli -t -f SIGNAL dev wifi"):read()
             wifi_now.connected = true
         end
--- a/widget/contrib/wifi.lua
+++ b/widget/contrib/wifi.lua
@@ -1,3 +1,5 @@
+local awful     = require("awful")
+local naughty   = require("naughty")
 local wibox     = require("wibox")
 local helpers   = require("lain.helpers")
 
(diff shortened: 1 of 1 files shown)


vlad-tdot:
Hah, I came here to contribute my widget that shows signal strength. What a timing

widget/contrib/wifi.lua
```diff
@@ -0,0 +1,45 @@
...
+local helpers   = require("lain.helpers")
+
+function factory(args)
+    local wifi      =  { widget = wibox.widget.textbox() }
+    local args      = args or {}
+    local timeout   = args.timeout or 2
+    local settings  = args.settings or function() end
+         
+    function wifi.is_connected()
+        conn_flag = io.popen("[[ ! -z `ping -c 1 -w 3 www.duckduckgo.com` ]] && echo true"):read()
```
lcpz: Never use `io.popen` in Awesome, because it's IO blocking.

Read [here](https://awesomewm.org/doc/api/libraries/awful.spawn.html).

widget/contrib/wifi.lua
```diff
@@ -0,0 +1,45 @@
...
+    function wifi.update()
+        wifi_now = {
+            ssid = "N/A",
+            signal = "/NA",
+            enable_signal = true,
+            connected = false
+        }
+
+        if wifi.is_connected() then
+            wifi_now.ssid = io.popen("nmcli -t -f SSID dev wifi list"):read()
```
lcpz: This introduces NetworkManager as dependency.

lcpz:
@amirulmenjeni Sorry, I have to reject this pull. See why in my comments.

Moreover, I already wrote a recipe you can find [here](https://github.com/lcpz/awesome-copycats/blob/master/themes/vertex/theme.lua#L230-L270), which shows signal strength, as @vlad-tdot wish.

There is little to do in order to show the SSID as well, and I could do it myself, but I'd like to make you Awesome first time contributor(s). Take my snippet, extend it with the features you want, assuring that you don't add unnecessary dependencies and always use Awesome facilities for asynchronous operations, and finally read [here](https://awesomewm.org/recipes/) to submit your recipe.

I'll leave this thread open as long as you need my support.

amirulmenjeni:
@lcpz Thank you. Do you consider nmcli from the networkmanager package to be an unnecessary dependency? I'd opt to use `/proc/net/wireless` file, but unfortunately that file on my system is not populated with data unless I read it using `sudo`. A bit of searching tells me it's something to do with my broadcom-wl driver. Anyway, I'll see what I can do.

lcpz:
@amirulmenjeni You should be able to get the informations without being a sudoer, just like `nmcli` does. If you can't come up with a solution, read `nmcli` code to see how it does it. Let me know.

amirulmenjeni:
@lcpz Looking up networkmanager source code, I think [this](https://github.com/NetworkManager/NetworkManager/blob/5b50564a8a9ab7b656800d6d33040add8d7791d9/src/platform/wifi/wifi-utils-wext.c#L328) is the source file of note. It looks like the program obtain the wifi data using `ioctl` in the `wifi_wext_get_qual` function, and parse the data in `wext_qual_to_percent` function.

I think it's ideal to use `/proc/net/wireless`. But this would mean system with bad driver like mine won't be able to see the signal strength. I'm also unable to use `iw*` commands to get the signal strength information as well, unless running the `iw*` commands with sudo. The SSID and devices can still be easily obtainable though.

lcpz:
Which OS are you using? Possibly the issue is not with the drivers, but just with system permissions.

amirulmenjeni:
@lcpz Linux arch 4.14.5-1-ARCH

lcpz:
I have the same.

Naive attempt: why not `sudo chmod +r /proc/net/wireless`? Or are its permissions being resetted?

amirulmenjeni:
@lcpz The permission is already set to `444. 

Another thing: I mentioned earlier that I think nmcli uses `ioctl` system call and pass `SIOCGIWSTATS` request to get the wifi data, but I now I think that this isn't the case. The output of `strace nmcli dev wifi` shows no indication of `ioctl` call that pass `SIOCGIWSTATS` request.

amirulmenjeni:
@lcpz Looks like NetworkManager is running as root afterall:

```
$ ps aux | grep NetworkManager

root       372  0.0  0.4 469236 16484 ?        Ssl  Dec15   0:02 /usr/bin/NetworkManager --no-daemon
amenji   29885  0.0  0.0  10888  2320 pts/3    S+   11:44   0:00 grep --color=auto NetworkManager

```
Thanks to the guy who answered [my question](https://unix.stackexchange.com/questions/411189/how-does-networkmanager-nmcli-able-to-get-my-wifi-signal-strength-without-root/411190#411190) for pointing this out.

What do you say of a fallback method, say, allowing the user to the pass commands as argument to the widget when taking the information from `/proc/net/wireless` is not possible? The stdout of the command can then be used to show the wifi detail .e.g. its strength in the case when `/proc/net/wireless` fails to show any useful data.

lcpz:
What about [this other answer](https://unix.stackexchange.com/a/411212) instead? Are you able to run

```shell
/sbin/iw wlan0 station dump
```

without being a sudoer?

If so, we can define 2 recipes, or include 2 different fetch methods in one.

amirulmenjeni:
@lcpz Running that command with or without being a sudoer doesn't print out any output, not even an error message.

lcpz:
Try with your device id.

```shell
$(which iw) wlp6s0 station dump
```

amirulmenjeni:
@lcpz `iw` does work on my machine. For example `iw wlp6s0 info` does show some information, but it doesn't tell me the signal strength. Same thing with `iwlist wlp6s0 scan`, it shows does shows the quality and signal level, but to get updated information I need to be root. I think this is strictly a driver issue, like someone in this [thead](https://bbs.archlinux.org/viewtopic.php?id=207674) had. I use the same driver as the poster. 

Using `$(which iw)` does the same thing as well, no luck there.

lcpz:
If this is really a poor driver bug, I'm afraid I can't do anything to solve it.

I can only help you writing a correct `nmcli` recipe, while I'll add SSID support to mine and do a pull request to Awesome.

amirulmenjeni:
Are you suggesting we use `/proc/net/wireless` and use `nmcli` as fallback method? And what do you mean about the pull request?

lcpz:
I mean I write the recipe with `/proc/net/wireless`, and help you writing a correct `nmcli` version.

We can push both to Awesome recipes webpage section linked above.

amirulmenjeni:
Alright, I'll write the `nmcli` version and push it here so you can see it.

widget/contrib/wifi.lua
```diff
@@ -0,0 +1,70 @@
...
+        signal = args.signal_placeholder or "N/A",
+    }
+
+    function wifi.shell_cmd(cmd)
+        return awful.util.shell .. " -c '" .. cmd .. "'"
+    end
+
+    function wifi.update()
+
+        awful.spawn.easy_async(wifi.shell_cmd(command.check_conn),
```
lcpz: Use [this](https://awesomewm.org/doc/api/libraries/awful.spawn.html#easy_async_with_shell) and remove `wifi.shell_cmd`.

widget/contrib/wifi.lua
```diff
@@ -0,0 +1,70 @@
...
+local wibox     = require("wibox")
+local helpers   = require("lain.helpers")
+
+function factory(args)
+    local wifi      =  { widget = wibox.widget.textbox() }
+    local args      = args or {}
+    local timeout   = args.timeout or 2
+    local settings  = args.settings or function() end
+         
+    local command = {
```
lcpz: Don't use `ping` to check connection. `nmcli` must have its own method. Maybe `nmcli general`.

widget/contrib/wifi.lua
```diff
@@ -0,0 +1,70 @@
...
+        check_conn = "[ ! -z \"`ping -c1 -w3 www.duckduckgo.com`\" ] && echo -n 1",
+        nmcli_dev = "nmcli -t dev | grep -E ^wl",
+        nmcli_wifi = "nmcli -g IN-USE,SSID,SIGNAL dev wifi list | grep -F \"*\"",
+    }
```
lcpz: Condense all in one command and one asynchronous operation, for instance:

```lua
local command = "nmcli general && nmcli -t dev | grep -E ^wl && nmcli -g IN-USE,SSID,SIGNAL dev wifi list | grep -F \"*\""
```

widget/contrib/wifi.lua
```diff
@@ -0,0 +1,70 @@
...
+
+    wifi_now = {
+        device = args.device or "N/A",
+        connected = false,
```
lcpz: Better would be `state = "UNKNOWN" or  "UP" or "DOWN"` like `ip link`.

widget/contrib/wifi.lua
```diff
@@ -0,0 +1,70 @@
...
+        ssid = args.ssid_placeholder or "N/A",
+        signal = args.signal_placeholder or "N/A",
+    }
+
+    function wifi.shell_cmd(cmd)
```
lcpz: Since you have to have just one asynchronous call, this function is useless.

widget/contrib/wifi.lua
lcpz: Use instead [this](https://awesomewm.org/doc/api/libraries/awful.spawn.html#easy_async_with_shell) function I pushed a while ago.

widget/contrib/wifi.lua
```diff
@@ -0,0 +1,70 @@
...
+            end
+
+            awful.spawn.easy_async(wifi.shell_cmd(command.nmcli_wifi),
+                function(stdout, stderr, reason, exit_code)
+                    local ssid, signal = string.match(stdout, ":(.-):(%d+)")
+                    wifi_now.ssid = ssid
+                    wifi_now.signal = signal
+                end
+            )
+        end
```
lcpz: As said above, merge all these operations in one.

widget/contrib/wifi.lua
```diff
@@ -0,0 +1,70 @@
...
+        widget = wifi.widget
+        settings()
+    end
+
+    helpers.newtimer("wifi", timeout, wifi.update)
+
+    return wifi
+end
+
+return factory
```
lcpz: Also, start to think how to turn this into a [watcher](https://awesomewm.org/recipes/watch/).

lcpz:
I want to ease your work. This is the final structure:

```lua
local nmcli_status = "nmcli general"
local nmcli_dev    = "nmcli -t dev"
local nmcli_wifi   = "nmcli -g IN-USE,SSID,SIGNAL dev wifi list"

local nmcli_widget = awful.widget.watch(
    { awful.util.shell, "-c", string.format("%s && %s && %s",
    nmcli_status, nmcli_dev, nmcli_wifi), 2,
    function(widget, stdout)
        local wifi_now = {}

        -- TODO
        wifi_now.device = stdout:match() or "N/A"
        wifi_now.status = stdout:match() or "N/A"
        wifi_now.ssid   = stdout:match() or "N/A"
        wifi_now.signal = stdout:match() or "N/A"

        -- customize here
        widget:set_text(wifi_now.ssid..": "..wifi_now.signal)
    end
)
```

Just do the `TODO` matches. Also, I believe you can merge the `nmcli` commands in one, but you have to figure it out by yourself, since I don't use NetworkManager.

amirulmenjeni:
Thank you. I now use
```
nmcli -g ACTIVE,IN-USE,DEVICE,SSID,SIGNAL dev wifi
```
which shows a list of available connections The top connection listed is the connection with the highest signal strength. The value in the ACTIVE column is used to imply the connection status: If there exists one connection in the list that have the ACTIVE column labeled as "yes", then it's connected. Otherwise, it's disconnected. 

```lua
local nmcli_widget = awful.widget.watch(                                            
    { awful.util.shell, "-c", "nmcli -g ACTIVE,IN-USE,DEVICE,SSID,SIGNAL dev wifi" }, 2,                                      
    function(widget, stdout)                                                        
        local wifi_now = {}                                                         
                                                                                    
        -- Wifi is connected if a there exist an active connection 
        -- represented by a "yes" in the ACTIVE column.                  
        -- Only one connection may be active at a time.                             
        if stdout:match("(yes): ?*?:") then                                         
            wifi_now.status = "connected"                                           
        else                                                                        
            wifi_now.status = "disconnected"                                             
        end                                                                         
                                                                                    
        -- If no connection is active, display the top connection             
        -- from the list, which have the highest signal strength. Otherwise,        
        -- show the information from the active connection.                         
        if wifi_now.status == "disconnected" then                                   
            wifi_now.device = stdout:match("^.-:.-:(.-):") or "N/A"                 
            wifi_now.ssid   = stdout:match("^.-:.-:.-:(.-):") or "N/A"              
            wifi_now.signal = stdout:match("^.-:.-:.-:.-:(%d+)") or "N/A"           
        else                                                                        
            wifi_now.device = stdout:match("yes:%*:(.-):") or "N/A"                 
            wifi_now.ssid   = stdout:match("yes:%*:.-:(.-):") or "N/A"              
            wifi_now.signal = stdout:match("yes:%*:.-:.-:(%d+)") or "N/A"           
        end                                                                         
                                                                                    
        -- customize here                                                           
        widget:set_text(wifi_now.status .. " " .. wifi_now.device .. " " .. wifi_now.ssid .. " " .. wifi_now.signal )
    end                                                                          
)  
```

lcpz:
OK, this is the final recipe.

```lua
local nmcli_widget = awful.widget.watch(
    { awful.util.shell, "-c", "nmcli -g ACTIVE,IN-USE,DEVICE,SSID,SIGNAL dev wifi" }, 2,
    function(widget, stdout)
        local wifi_now = {}

        -- Wifi is connected if a there exist an active connection
        -- represented by a "yes" in the ACTIVE column.
        -- Only one connection may be active at a time.
        wifi_now.status = stdout:match("(yes): ?*?:") and "connected" or "disconnected"

        -- If no connection is active, display the top connection
        -- from the list, which have the highest signal strength.
        -- Otherwise, show the information from the active connection.
        if wifi_now.status == "disconnected" then
            wifi_now.device = stdout:match("^.-:.-:(.-):") or "N/A"
            wifi_now.ssid   = stdout:match("^.-:.-:.-:(.-):") or "N/A"
            wifi_now.signal = stdout:match("^.-:.-:.-:.-:(%d+)") or "N/A"
        else
            wifi_now.device = stdout:match("yes:%*:(.-):") or "N/A"
            wifi_now.ssid   = stdout:match("yes:%*:.-:(.-):") or "N/A"
            wifi_now.signal = stdout:match("yes:%*:.-:.-:(%d+)") or "N/A"
        end

        -- customize here
        widget:set_text(wifi_now.status.." "..wifi_now.device.." "..
                        wifi_now.ssid.." "..wifi_now.signal)
    end
)
```

You can create a pull request to [awesome-www](https://github.com/awesomeWM/awesome-www). Take [this PR](https://github.com/awesomeWM/awesome-www/pull/108) as an example on how to do.

Closing.

PR #373, opened 2017-12-14
Wifi signal strength widget

vlad-tdot:


vlad-tdot:
Just went through the comments on the other pull request and realized that there's already a recipe for this, so nevermind

lcpz:
See if you can contribute as requested in #372.

PR #376, opened 2017-12-17
Updates widget

alfunx:
A new widget that will show the number of available updates. Command to check updates can be configured or chosen from the provided ones. Will show a notification when the update count has increased (can be deactivated). This widget may be useful for people who use rolling release distributions, perhaps less useful for point release distribution users.

With the provided options you could for example set up two separate widgets, for `pacman` and `pacaur`, or a combined one.

Updates widget
An example; note the pacman icon.

I tried to keep this as general as possible, so it is possible to show outdated `pip` packages with `pip list --outdated --format=legacy` for example. You could use it for anything unrelated too, as what it basically does is line-counting and notifying when the line count changes.

Usage may look like this:

```lua
local pacman_widget = lain.widget.contrib.updates({
    package_manager = "pacman",
    notify = true,
    settings = function()
        widget:set_markup(markup.fontfg(normal_font, normal_color, available))
    end
})
```

### DIFF
--- a/widget/contrib/updates.lua
+++ b/widget/contrib/updates.lua
@@ -1,0 +1,84 @@
+--[[
+
+     Licensed under GNU General Public License v2
+      * (c) 2017, Alphonse Mariyagnanaseelan
+
+--]]
+
+local helpers       = require("lain.helpers")
+local awful         = require("awful")
+local shell         = require("awful.util").shell
+local wibox         = require("wibox")
+local naughty       = require("naughty")
+local lines, floor  = io.lines, math.floor
+local string        = { format = string.format,
+                        gsub   = string.gsub,
+                        len    = string.len}
+
+-- Available updates
+-- lain.widget.contrib.updates
+
+local function factory(args)
+    local updates          = { widget = wibox.widget.textbox() }
+    local args            = args or {}
+    local timeout         = args.timeout or 900
+
+    local notify          = args.notify or false
+    local notify_title    = args.notify_title or "Updates"
+    local notify_timeout  = args.notify_timeout or 15
+
+    local settings        = args.settings or function() end
+    local package_manager = args.package_manager
+    local command         = args.command or ""
+
+    local commands = {
+        pacman        = "checkupdates | sed 's/->/→/' | column -t",
+        pacaur        = "pacaur -k --color never | sed 's/:: [a-zA-Z0-9]\\+ //' | sed 's/->/→/' | column -t",
+        pacman_pacaur = "( checkupdates & pacaur -k --color never | sed 's/:: [a-zA-Z0-9]\\+ //' ) | sed 's/->/→/' | sort | column -t",
+        dnf           = "dnf check-update --quiet",
+        apt           = "apt-show-versions -u"
+    }
+
+    local update_count = 0
+
+    function updates.update(notify)
+        helpers.async({ shell, "-c", commands[package_manager] or command }, function(update_text)
+            available  = tonumber((update_text:gsub('[^\n]', '')):len())
+            widget = updates.widget
+
+            if available > update_count and notify then
+                updates.show_notification(update_text)
+            end
+            update_count = available
+            settings()
+        end)
+    end
+
+    function updates.automatic_update()
+        -- Notify if set and update_count has increased
+        updates.update(notify)
+    end
+
+    function updates.manual_update()
+        -- Allways show notification
+        update_count = -1
+        updates.update(true)
+    end
+
+    function updates.show_notification(update_text)
+        if not update_text or update_text == "" then
+            update_text = "None."
+        end
+        naughty.notify({
+            title = notify_title,
+            text = string.gsub(update_text, '[\n%s]*$', ''),
+            timeout = notify_timeout
+        })
+    end
+
+    helpers.newtimer("updates", timeout, updates.automatic_update)
+
+    return updates
+end
+
+return factory
--- a/widget/contrib/updates.lua
+++ b/widget/contrib/updates.lua
@@ -19,34 +19,37 @@ local string        = { format = string.format,
 -- lain.widget.contrib.updates
 
 local function factory(args)
-    local updates          = { widget = wibox.widget.textbox() }
+    local updates         = { widget = wibox.widget.textbox() }
     local args            = args or {}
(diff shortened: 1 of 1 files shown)


widget/contrib/updates.lua
```diff
@@ -0,0 +1,84 @@
...
+-- Available updates
+-- lain.widget.contrib.updates
+
+local function factory(args)
+    local updates          = { widget = wibox.widget.textbox() }
+    local args            = args or {}
+    local timeout         = args.timeout or 900
+
+    local notify          = args.notify or false
+    local notify_title    = args.notify_title or "Updates"
```
lcpz: Use a table like [this one](https://github.com/lcpz/lain/blob/master/widget/calendar.lua#L114) instead of setting a variable for each parameter.

widget/contrib/updates.lua
```diff
@@ -0,0 +1,84 @@
...
+
+-- Available updates
+-- lain.widget.contrib.updates
+
+local function factory(args)
+    local updates          = { widget = wibox.widget.textbox() }
+    local args            = args or {}
+    local timeout         = args.timeout or 900
+
+    local notify          = args.notify or false
```
lcpz: Use flag strings instead, like [here](https://github.com/lcpz/lain/blob/master/widget/fs.lua#L48).

widget/contrib/updates.lua
```diff
@@ -0,0 +1,84 @@
...
+    local notify_title    = args.notify_title or "Updates"
+    local notify_timeout  = args.notify_timeout or 15
+
+    local settings        = args.settings or function() end
+    local package_manager = args.package_manager
+    local command         = args.command or ""
+
+    local commands = {
+        pacman        = "checkupdates | sed 's/->/→/' | column -t",
```
lcpz: What is this `checkupdates`? Remember not to add custom scripts calls or new dependencies.

widget/contrib/updates.lua
```diff
@@ -0,0 +1,84 @@
...
+        pacaur        = "pacaur -k --color never | sed 's/:: [a-zA-Z0-9]\\+ //' | sed 's/->/→/' | column -t",
+        pacman_pacaur = "( checkupdates & pacaur -k --color never | sed 's/:: [a-zA-Z0-9]\\+ //' ) | sed 's/->/→/' | sort | column -t",
+        dnf           = "dnf check-update --quiet",
+        apt           = "apt-show-versions -u"
+    }
```
lcpz: I am not sure about this command table either. Potentially, since it should be as comprehensive as possible, it could grow in size without limits. Better would be to specify the command as argument, and to provide some examples in the wiki entry.

widget/contrib/updates.lua
```diff
@@ -0,0 +1,84 @@
...
+
+            if available > update_count and notify then
+                updates.show_notification(update_text)
+            end
+            update_count = available
+            settings()
+        end)
+    end
+
+    function updates.automatic_update()
```
lcpz: Pointless.

widget/contrib/updates.lua
```diff
@@ -0,0 +1,84 @@
...
+            update_text = "None."
+        end
+        naughty.notify({
+            title = notify_title,
+            text = string.gsub(update_text, '[\n%s]*$', ''),
+            timeout = notify_timeout
+        })
+    end
+
+    helpers.newtimer("updates", timeout, updates.automatic_update)
```
lcpz:
```diff
-    helpers.newtimer("updates", timeout, updates.automatic_update)
+    helpers.newtimer("updates", timeout, updates.update(notify)) -- if you keep the argument
```

widget/contrib/updates.lua
```diff
@@ -0,0 +1,84 @@
...
+        updates.update(notify)
+    end
+
+    function updates.manual_update()
+        -- Allways show notification
+        update_count = -1
+        updates.update(true)
+    end
+
+    function updates.show_notification(update_text)
```
lcpz: Can be merged into `.update`. No need for a separate function which is used only by another function.

widget/contrib/updates.lua
```diff
@@ -0,0 +1,84 @@
...
+        if not update_text or update_text == "" then
+            update_text = "None."
+        end
+        naughty.notify({
```
lcpz: What about `naughty.destroy` and duplicate notifications? See how `naughty` is managed in [`calendar.lua`](https://github.com/lcpz/lain/blob/master/widget/calendar.lua).

lcpz:
There's a lot to improve. See my comments.

I believe this can be another [recipe](https://awesomewm.org/recipes/).

alfunx:
Thanks! Will look into it.

widget/contrib/updates.lua
alfunx: `checkupdates` is a script provided by `pacman`, the Arch package manager.

widget/contrib/updates.lua
alfunx: That was something I considered for convenience. But you're right, other than that potential users will most likely figure this out.

alfunx:
I couldn't apply your last suggestion. AFAIK you can't pass a function with an argument already set, or did I misunderstand something?

lcpz:
I'll spare you some work. Here's a simple recipe.

```lua
local upcmd = "checkupdates | sed 's/->/→/' | column -t" -- pacman
--local upcmd = "dnf check-update --quiet"                 -- dnf
--local upcmd = "apt-show-versions -u"                     -- apt
--local upcmd = "pip list --outdated --format=legacy"      -- pip

local upcount, upnotify = 0, "on"

local updates = awful.widget.watch(
    { awful.util.shell, "-c", upcmd }, 900,
    function(widget, stdout)
        available = stdout:gsub("[^\n]", ""):len()

        if available > upcount and upnotify == "on" then
            -- customize here
            naughty.notify {
                text = update_text:gsub("[\n%s]*$", "")
            }
        end

        upcount = available

        -- customize here
        widget:set_text("Updates: "..available)
    end
)
```

Correct it and create a pull request to [awesome-www](https://github.com/awesomeWM/awesome-www). Take [this PR](https://github.com/awesomeWM/awesome-www/pull/108) as an example on how to do.

Closing.

PR #377, opened 2017-12-30
menu_iterator: naughty.notify based menu utility

sim590:
This is a generic menu utility used for e.g. on a [project][myproject] I have just created (see [the file][menu_iterator]). 

The code is adapted from a less generic version in the lua script for creating a menu for choosing monitor layouts with [xrandr][]. This script is advertised on the [wiki][]. Shouldn't the Lain project integrate this script? IMO, it's a must widget.

Assuming both this PR and the xrandr script would be integrated, you should consider integrating an adapted version of the xrandr script so that it uses the generic menu iterator. An adapted version can be found [here][adapted-xrandr] in my dotfiles.

[myproject]: https://github.com/sim590/amh
[menu_iterator]: https://github.com/sim590/amh/blob/master/util/menu_iterator.lua
[xrandr]: https://github.com/awesomeWM/awesome-www/blob/master/recipes/xrandr.lua
[wiki]: https://awesomewm.org/recipes/xrandr/
[adapted-xrandr]: https://github.com/sim590/dotfiles/blob/53e25b1d4ca294aded03161346509453f457945f/awesome/rc/xrandr.lua


sim590:
I have just pushed some new commits. Those changes help a lot for the user's convenience. Can you comment please?

helpers.lua
```diff
@@ -170,6 +170,27 @@ function helpers.spairs(t)
     end
 end
 
+-- create trivial partition of a set
+function helpers.trivial_partition_set(set)
```
lcpz: I'd rather call it `partition_of_singletons` or add the descriptions from the commit message to the code.

util/menu_iterator.lua
```diff
@@ -62,7 +63,72 @@ local function iterate(menu, timeout, icon)
...
+--                    of possible hosts, but not selected). Takes the host as a
+--                    string argument.
+--   * extra_choices: an array of pairs { choice_text, cb } for extra choices to
+--                    be added to the menu.
+--   * combination:   the combination of host to generate. Possible choices are
+--                    "powerset" and "single" (the default).
+local function menu(args)
+    local hosts       = assert(args.hosts or args[1])
+    local name        = assert(args.name or args[2])
+    local selected_cb = args.selected_cb
```
lcpz: What if `args` doesn't have `selected_cb`? Put a fallback, for instance:

```lua
local selected_cb = args.selected_cb or function() end
```

util/menu_iterator.lua
```diff
@@ -62,7 +63,72 @@ local function iterate(menu, timeout, icon)
...
+    local rejected_cb = args.rejected_cb
```
lcpz: Like above.

util/menu_iterator.lua
```diff
@@ -62,7 +63,72 @@ local function iterate(menu, timeout, icon)
...
+
+    return m
+end
+
+return {
+    iterate = iterate,
+    menu    = menu
+}
 
 -- vim:set et sw=4 ts=4 tw=120:
```
lcpz: Remove extra spaces and vim options.

lcpz:
I don't merge recipes, I actually [do the contrary](https://github.com/awesomeWM/awesome-www/pull/108).

Please give me some concrete examples or use cases in order to clearly understand why this should be a compulsory widget. I see you use this for monitors and hosts, but you can make your point more clear.

sim590:
I have just realized that the comments of the `menu` function are those taken from [amh](https://github.com/sim590/amh/blob/master/util/init.lua#L148). I should have replaced those with some more accurate and generic.

- [ ] Comment `menu_iterator.menu` more appropriately.

> I don't merge recipes, I actually do the contrary.

I don't really understand what you mean by "recipe".

> Please give me some concrete examples or use cases in order to understand why this should be a compulsory widget. I see you use this for monitors and hosts, but you can make your point more clear.

May be I placed the code in the wrong file. Tell me if it should be elsewhere. This is supposed to be a generic menu utility letting people chose actions upon user choice for any use cases really. Some examples:

- Chose between different `xrandr` configurations  like I suggested in my initial comment;
- Quickly switch to different modes: for e.g., recording screen 0, then switching to recording screen 1 in a screencast;
- Changing behaviour of a subset of keymaps: iterating over all keymap modes would let the user choose a mode quickly;

Possibilities are really numerous and this tool makes the use cases' experience convenient. I think that this tool is really generic and can be used for a vast number of things we don't know of and that's why I think it belongs in a widely known awesome extension such as lain which provide helpers and utilities among other things like widgets, indeed.

util/menu_iterator.lua
sim590: This is actually checked [here](https://github.com/lcpz/lain/blob/9e21a95a1a20a929a4a0f62f3ffddeeb6e041045/util/menu_iterator.lua#L22). Empty callbacks are used, for instance like [this][example].

[example]: https://github.com/sim590/amh/commit/c648aa927649e8c5388362e4104afe04be672d48#diff-42b565af082eb6b5eca071ecb51c0225R36

util/menu_iterator.lua
sim590: *cf* [my response above,](https://github.com/lcpz/lain/pull/377#discussion_r168613952).

lcpz:
> I don't really understand what you mean by "recipe".

As you can see in the link above, a *recipe* is a self-contained snippet of code which provides some functionality. In particular, here a recipe is a widget written with [`awful.widget.watch`](https://awesomewm.org/recipes/watch/) to replace an existing Lain counterpart. The aim is to lighten Lain and make the same functionalities supported by the WM in a future-proof fashion.

> I think that this tool is really generic and can be used for a vast number of things we don't know of [...]

Can you provide some *examples* (actual pieces of code) which show how these menu utilities can be used? Since you're using them yourself, you can just extract and explain/comment something from your configurations.

util/menu_iterator.lua
lcpz: This indicates that there are things you didn't explain. You have to do it, either by adding comments or writing a proper wiki entry, in which you should also put some usage examples (see [here](https://github.com/lcpz/lain/pull/377#issuecomment-366209236)).

sim590:
You can see example usage and gifs on the [amh home page][home-page]. For instance, I use it for selecting modes of configuration for [synergy][]:

```lua
util.menu({
    hosts       = hosts,
    name        = "Synergy",
    selected_cb = function(host) synergy(host) end,
    rejected_cb = function(host) util.remote_spawn(host, "pkill -u $USER -x synergyc", nil, false) end,
    extra_choices = { { "None", nil } },
    combination = "powerset"
})
```

The call to the function `menu` will return a menu compatible with the function `iterate` in order to iterate on it using the `naughty.notify` utility. To do so, one has to write something like [this][synergy-iterate] example where I iterate over the menu for synergy which is created above:

```lua
util.menu_iterator.iterate(synergy.menu(amh.hosts), 4, amh.synergy_icon)
```

<p align="center"></p>

Subsequent calls to the `iterate` function will step further into the list of choice to actually iterate over the list.  Once the timeout has passed (4 seconds by default) without anymore calls to `iterate`, the choice is made and it calls the associated callbacks (for selected and rejected choices) configured in the menu.

[synergy-iterate]: https://github.com/sim590/amh/blob/master/init.lua#L32
[synergy]: https://github.com/sim590/amh/blob/master/exec/synergy.lua
[home-page]: https://github.com/sim590/amh/

lcpz:
Your case is very specific, it'll be better if you provide an example where no external or third party elements are used.

Anyway, I have no problems in merging. Once you do the requested changes (see my comments), I'll review again. Don't forget to add a wiki entry with the above infos.

sim590:
> Anyway, I have no problems in merging. Once you do the requested changes (see my comments), I'll review again. Don't forget to add a wiki entry with the above infos.

I don't see anymore comments other than those that I have made changes for before rebasing my branch and force pushing already. Is it not updated already?

> Your case is very specific, it'll be better if you provide an example where no external or third party elements are used.

I will surely provide a relevant and generic enough example in the wiki as you suggested so that everything is clear without making references to specific and out of context knowledge.

- [ ] provide wiki article, open issue about wiki page

I'll make changes to the wiki locally and open an issue about it for you to review my changes and merge them in. I think that it is a fine workflow which will permit better control on the wiki integration as suggested by this [thread](https://stackoverflow.com/questions/10642928/how-to-pull-request-a-wiki-page-on-github#11481887)

lcpz:
> Is it not updated already?

As far as I can see here, no.

Assure that you pushed, or try to squash and rebase again in one commit.

sim590:
I saw "outdated diff", then didn't catch that you had commented further on that diff. I unfolded and saw. From what I understand, you want me to comment in the code about where it is checked and that passing nil is part of the interface of the function. Is that right?

updated

Otherwise, all have been answered:

> I'd rather call it partition_of_singletons or add the descriptions from the commit message to the code.

I added the comment.

> Remove extra spaces and vim options.

It was removed

util/menu_iterator.lua
```diff
@@ -63,4 +64,76 @@ local function iterate(menu, timeout, icon)
...
+--   choices (the powerset)
+--
+-- The following describes the function arguments:
+-- * args: an array containing the following members:
+--   * choices:       the list of choices from which to generate the menu
+--   * name:          the displayed name of the menu (in the form "name: choices")
+--   * selected_cb:   the callback to execute for each selected choice. Takes
+--                    the choice as a string argument. The function
+--                    menu_iterator.naughty_destroy_callback will handle nil
+--                    callbacks. It is then fine to pass nil callbacks.
```
sim590: @lcpz Here I comment about the callbacks usage.

util/menu_iterator.lua
```diff
@@ -0,0 +1,66 @@
...
+
+local state = { cid = nil }
+
+local function naughty_destroy_callback(reason)
+    if reason == naughty.notificationClosedReason.expired or
+        reason == naughty.notificationClosedReason.dismissedByUser then
+        local actions = state.index and state.menu[state.index - 1][2]
+        if actions then
+            for _,action in pairs(actions) do
+                -- don't try to call nil callbacks
```
sim590: I added another comment here just to be sure even though the code below literally says what the comment talks about.

PR #380, opened 2018-01-21
Extra functionality in net.lua, full battery notification, a new widget for monitoring external drives

cool-cool-sweat:
The new helper function helpers.line_callback gets used in net.lua so that all network interfaces except loopback ones get added to net.iface, even if they don't have an active connection. This makes the widget more flexible since it can now detect changes in the interface after the widget's been loaded. This is useful if for example the widget's loaded while the ethernet's unconnected and then later the user switches to ethernet.

helpers.async and helpers.async_with_shell can now return the process' exit code, so widgets can use it in if statements. extdrive.lua makes use of that

### DIFF
--- a/widget/bat.lua
+++ b/widget/bat.lua
@@ -47,6 +47,14 @@ local function factory(args)
         bg      = "#CDCDCD"
     }
 
+    bat_notification_charged_preset = {
+        title = "Battery full",
+        text  = "You can unplug the cable",
+        timeout = 15,
+        fg      = "#202020",
+        bg      = "#CDCDCD"
+    }
+
     bat_now = {
         status    = "N/A",
         ac_status = "N/A",
@@ -156,16 +164,23 @@ local function factory(args)
         widget = bat.widget
         settings()
 
-        -- notifications for critical and low levels
-        if notify == "on" and bat_now.status == "Discharging" then
-            if tonumber(bat_now.perc) <= n_perc[1] then
+        -- notifications for critical, low, and full levels
+        if notify == "on" then
+            if bat_now.status == "Discharging" then
+                if tonumber(bat_now.perc) <= n_perc[1] then
+                    bat.id = naughty.notify({
+                        preset = bat_notification_critical_preset,
+			            replaces_id = bat.id
+                    }).id
+                elseif tonumber(bat_now.perc) <= n_perc[2] then
+                    bat.id = naughty.notify({
+                        preset = bat_notification_low_preset,
+                        replaces_id = bat.id
+                    }).id
+                end
+            elseif bat_now.status == "Full" then
                 bat.id = naughty.notify({
-                    preset = bat_notification_critical_preset,
-                    replaces_id = bat.id
-                }).id
-            elseif tonumber(bat_now.perc) <= n_perc[2] then
-                bat.id = naughty.notify({
-                    preset = bat_notification_low_preset,
+                    preset = bat_notification_charged_preset,
                     replaces_id = bat.id
                 }).id
             end
--- a/widget/net.lua
+++ b/widget/net.lua
@@ -15,22 +15,23 @@ local string  = { format = string.format, match = string.match }
 -- lain.widget.net
 
 local function factory(args)
-    local net      = { widget = wibox.widget.textbox(), devices = {} }
-    local args     = args or {}
-    local timeout  = args.timeout or 2
-    local units    = args.units or 1024 -- KB
-    local notify   = args.notify or "on"
-    local screen   = args.screen or 1
-    local settings = args.settings or function() end
+    local net        = { widget = wibox.widget.textbox(), devices = {} }
+    local args       = args or {}
+    local timeout    = args.timeout or 2
+    local units      = args.units or 1024 -- KB
+    local notify     = args.notify or "on"
+    local wifi_state = args.wifi_state or "off"
+    local eth_state  = args.eth_state or "off"
+    local screen     = args.screen or 1
+    local settings   = args.settings or function() end
 
     -- Compatibility with old API where iface was a string corresponding to 1 interface
     net.iface = (args.iface and (type(args.iface) == "string" and {args.iface}) or
                 (type(args.iface) == "table" and args.iface)) or {}
 
     function net.get_device()
-        helpers.async(string.format("ip link show", device_cmd), function(ws)
-            ws = ws:match("(%w+): <BROADCAST,MULTICAST,.-UP,LOWER_UP>")
-            net.iface = ws and { ws } or {}
+        helpers.line_callback("ip link", function(line)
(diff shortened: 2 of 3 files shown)


widget/net.lua
```diff
@@ -9,28 +9,30 @@
...
-    local notify   = args.notify or "on"
-    local screen   = args.screen or 1
-    local settings = args.settings or function() end
+    local net       = { widget = wibox.widget.textbox(), devices = {} }
+    local args      = args or {}
+    local timeout   = args.timeout or 2
+    local units     = args.units or 1024 -- KB
+    local notify    = args.notify or "on"
+    local wifi_info = args.wifi_info or "off"
+    local eth_info  = args.eth_info or "off"
```
lcpz: The variables `{wifi,eth}_info` needs to be documented in the wiki. The notes you wrote in this commit description should also be reported.

widget/net.lua
```diff
@@ -66,9 +68,56 @@ local function factory(args)
             dev_now.last_t   = now_t
             dev_now.last_r   = now_r
 
+	    if wifi_info == "on" and os.execute(string.format("ls /sys/class/net/%s/wireless", dev)) and string.match(dev_now.carrier, "1") then
```
lcpz: Indent here.

widget/net.lua
```diff
@@ -66,9 +68,56 @@ local function factory(args)
...
+            dev_now.wifi = true
+
+            helpers.line_callback(string.format("iw dev %s link", dev), function(line)
```
lcpz: This process is being respawned at each `net.update()` timeout. Since it is asynchronous, it will likely cause a memory leak.

widget/net.lua
```diff
@@ -66,9 +68,56 @@ local function factory(args)
...
+                if bitrate_parse then bitrate = bitrate_parse end
+            end)
+
+            dev_now.essid   = essid
+            dev_now.freq    = freq
+            dev_now.signal  = signal
+            dev_now.bitrate = bitrate
+        end
+
+	    if eth_info == "on" and not os.execute(string.format("ls /sys/class/net/%s/wireless", dev)) and string.match(dev_now.carrier, "1") then
```
lcpz: Indent here.

widget/net.lua
```diff
@@ -66,9 +68,56 @@ local function factory(args)
...
+            dev_now.ethernet = true
+            dev_now.bitrate  = tonumber(helpers.first_line(string.format("/sys/class/net/%s/speed", dev)))
+	    end
+
+	    if (eth_info == "on" or wifi_info == "on") and string.match(dev_now.carrier, "1") then
+            helpers.line_callback("ip neigh", function(line)
```
lcpz: See line 74.

widget/net.lua
```diff
@@ -66,9 +68,56 @@ local function factory(args)
...
+                local net_mac_parse = string.match(line, dev .. " lladdr ([%w:]+)")
+                if net_mac_parse then net_mac = net_mac_parse end
+            end)
+
+            helpers.line_callback("ip addr show " .. dev, function(line)
```
lcpz: See line 74.

helpers.lua
```diff
@@ -5,7 +5,7 @@
 
 --]]
 
-local spawn = require("awful.spawn")
+local spawn      = require("awful.spawn")
```
lcpz: This is a good commit. If you separate it from this PR, I can merge it.

widget/bat.lua
```diff
@@ -32,19 +32,19 @@ local function factory(args)
     local settings  = args.settings or function() end
 
     bat_notification_critical_preset = {
-        title   = "Battery exhausted",
-        text    = "Shutdown imminent",
-        timeout = 15,
-        fg      = "#000000",
-        bg      = "#FFFFFF"
+        preset = naughty.config.presets.critical,
```
lcpz: For double safety, please check if this won't modify `naughty.config.presets.critical`.

lcpz:
There seems to be a major flaw in your logic. The flexibility you want to introduce in the net widget would require a respawn of the same processes at each update timeout. This is likely to led to a memory saturation. You can probably verify it by just monitoring `top` output.

Overall, I want these widgets to be light and simple, but there are some elements I can directly merge, and others we can work on to get as close to your desideratum as possible. Please, read the review.

helpers.lua
cool-cool-sweat: Just the change to line 8 or you mean all the changes to the file?

widget/bat.lua
cool-cool-sweat: It doesn't, but this line doesn't seem to actually affect the notification's appearance at least for me. If you want I'll change it back to the old style or I could manually add  naughty.config.preset.critical's settings

cool-cool-sweat:
I can't find a memory leak in top, and system memory use stays basically the same with the widget removed and with it working with `wifi_info` set to on. And I've used the widget like this all day without it gobbling memory. But if you prefer I'll undo the changes to net.lua

helpers.lua
lcpz: All changes to the file.

widget/bat.lua
lcpz: > this line doesn't seem to actually affect the notification's appearance at least for me

Are you referring to the added line? If so, what's the purpose of it?

cool-cool-sweat: The purpose was to apply the settings in naughty.config.presets.critical [https://awesomewm.org/doc/api/libraries/naughty.html#config.presets] to the notification, but that doesn't seem to actually work inside the widget

cool-cool-sweat:
Don't worry about it, I'm sure you have other things you'd like to do. I'll work on this some more in a bit. Thanks for your patience

lcpz:
Taking this repository to release 1.0 after 4+ years of git is surely a thing I want to do, so I'll definitely come back at this.

Thank you for your interest and work, it is appreciated. 😹

cool-cool-sweat:
I reverted the changes to the notification presets in bat.lua since that wasn't really working out and I changed net.lua. Now it just sees if an ethernet connection's active and doesn't gather any info about it, and the only info it gets about the wifi's the signal strength. I have a draft edit for net.lua's wiki page

lcpz:
I've merged all except 7e9700f. I believe you can just use a [watcher](https://awesomewm.org/doc/api/classes/awful.widget.watch.html) for that. Example snippet:
```lua
local extdrive = awful.widget.watch(
    { awful.util.shell, "-c", "findmnt -n -o TARGET /dev/sdb1" }, 5,
    function(widget, stdout)
        -- do something to markup widget
    end
)
```
Once you've done, you can PR your recipe to [awesome-www](https://github.com/awesomeWM/awesome-www).

Don't forget to update the wiki. Thank you for your work. 👍

PR #383, opened 2018-02-12
Fix 370

trap000d:
https://github.com/lcpz/lain/issues/370
After week of testing on 3 computers seems it shows day/night status properly


lcpz:
Rebase into a single commit, please.

Also, be sure to put some comments explaining the code.

PR #388, opened 2018-02-22
fix for quake buginess

nonchip:
sometimes, `self:display` wouldn't return a client object, so calling `move_to_tag` on its return value directly would throw errors. this fixes that by inserting a simple `if` statement to check the return value.


lcpz:
Thank you, but please write a useful commit message next time.

nonchip:
sorry, I first just lazily committed in my own local clone before deciding to fork+PR it, will keep that in mind for the future

PR #404, opened 2018-10-26
Added more customizations

mtxr:
- Added margin and padding customizations.
- Added the ability to scroll over the widget to increase and decrease the volume.

### DIFF
--- a/widget/alsabar.lua
+++ b/widget/alsabar.lua
@@ -36,8 +36,11 @@ local function factory(args)
     local settings   = args.settings or function() end
     local width      = args.width or 63
     local height     = args.height or 1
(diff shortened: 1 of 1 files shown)


lcpz:
Thank you for your interest.

You are submitting a change for `alsabar`, but you should also do the same for `pulsebar`. The wiki should be updated as well. Since you are a first-time contributor, I will do it for you on this round.

Note that you could also do this and other customisations on the bar itself, calling it with `alsabar.bar`.

lcpz:
Actually, I forgot that buttons/keybindings have to be [optional](https://github.com/lcpz/lain/wiki/alsabar#buttons). Reverting that.

PR #416, opened 2018-12-01
Fix the new tp_smapi widget

BarbUk:


lcpz:
The `apipath` was also not correct. Fixed now.

PR #420, opened 2019-02-12
util/menu_iterator: retro compatibility with lua 5.1

sim590:
Since lua 5.2, `unpack` is no longer part of the global scope, but is now in `table`. This fix takes that into account in a retro-compatible fashion.

### DIFF
--- a/helpers.lua
+++ b/helpers.lua
@@ -12,7 +12,8 @@ local io         = { lines = io.lines,
                      open  = io.open }
 local pairs      = pairs
 local rawget     = rawget
-local table      = { sort  = table.sort }
+local table      = { sort  = table.sort, unpack = table.unpack }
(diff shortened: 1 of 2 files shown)


sim590:
I'm presently testing this. I'll remove the "WIP:" prefix when it's ready.

lcpz:
Please squash.

sim590:
I will if you insist, but each of the commits fix separate issues. The first is the retro compatibility issue I mentioned and the second is that `helpers.lua` was not `require`ed from `menu_iterator.lua` in the first place while it should have been. Do you still want to squash and hide this information? I really don't mind. I'm mentioning it in case you missed it.

lcpz:
The second commit is minimal, therefore yes, do as I ask. Thank you.

sim590:
It's done! :wink:

PR #421, opened 2019-02-19
Show week number in calendar widget

MethodicalMarmot:
Created an option in cal widget to show the week number before week row.


methodical-marmot:
image

lcpz:
Rendering is a bit ugly. Why not moving the week column to the right, with a header and no 0s for weeks < 10?

lcpz:
Merged in 012c6b6.

Issue #428, opened 2019-05-10
Weather Icons Unnecessary timezone convesrion(?)


AzureTwilight:
I am not sure about it but it seems that openweather API's have changed. The icon from returned json already has the correct day/night suffix (in EST).

I searched a bit but didn't find mentions on such change but I do need to comment out the gsub line to get correct weather icons. (https://github.com/lcpz/lain/blob/724c7b9a3a27e5de2077f388832167ae91e9f29c/widget/weather.lua#L135-L139)

Could anyone verify this?

trap000d:
Hi,
I'll test this (as author of this kludge :)
Indeed, I've noticed of wrong icon too several days ago on my work desktop, but had no opportunity to perform debug yet - full test takes 24 hours (there are several combinations of local, DST and UTC times which may cause accidental jumps to next or previous date).

trap000d:
Well. After testing on 2 different devices for 24 hours seems all looks good. I've just created PR.

PR #430, opened 2019-05-12
Weather status icon fix

trap000d:
In regarding to #428.
It looks like OpenWeatherMap have fixed most of issues in their API. Now they return sunrise and sunset times properly for current time zone, however the status icon is still wrong:
- transition N->D appears at 0:00 UTC (must be at local sunrise time)
- whereas D->N appears at local sunset time

So we still need to check if we're in or out from sunrise/sunset times, but no need to perform all other manipulations with dates and times.


lcpz:
Thank you.

PR #431, opened 2019-05-26
Find tempfile for x86 cpu if no tempfile is set

wooparadog:
The default thermal zone is usually wrong. 

This patch will look for the cpu zone by:

`/sys/class/thermal/zonex/type` == `x86_pkg_temp`


lcpz:
There are a couple of wrong things:

1. you are using `io.popen`, which is evil;
2. you are assuming that the architecture is always x86(_64).

Based on [this](https://github.com/lcpz/lain/issues/84#issuecomment-72751763), I rewrote the widget.

PR #438, opened 2019-08-10
alsabar: Calculate tot based on notification width

sprnza:
Currently tot is calculated on wibox width/height for some reason. Here is a proposal to do it based on notification width not ideal but still working much better as the old way. Another approach could be to add tot to the notification_preset table and let users to choose it manually.


lcpz:
I cannot merge because you are introducing `beautiful` as a dependency: this would break `awesome-copycats` and any configuration that relies on independent modules.

The reason for calculating `tot` based on wibox width/height is to tentatively make it scalable with the user resolution.

I will rename `tot` to `max_ticks` and make it an argument in `notification_preset`.

PR #439, opened 2019-08-10
add configuration arguments for pulsebar notification characters

gregflynn:
image

Configuration args for the pulsebar notification, didn't put a ton of thought into the names, open to suggestions


gregflynn:
image
with the 4th arg in the second commit

PR #457, opened 2020-06-10
Fix notification text for calendar widget

quizzmaster:
Since the `notification_preset` isn't apparently working for the notification text, just set the field `text` on it's own.


last5bits:
@quizzmaster Thank you, this fixed the problem for me.

@lcpz Could you please take a look? Thanks.

PR #471, opened 2020-11-14
Run fs widget update asynchronously

### LINKED ISSUE #462: Filesystem (FS) widget freezes Awesome due to remote filesystem (network query)
I'm using a few widgets from the lain library and it works great for a few years.
However, I have recently set up autofs to mount my remote filesystems (samba share and NFS) automatically. When there are issues/changes with the network, the filesystem (fs) widget often freezes the whole awesome wm (and it may return responsible after a while, probably because the query timed out).

The reason I say it's **the fs widget** that freezes the whole **awesome wm** is because:
1. If I remove the fs widget (both from the panel and its initialization), this freeze won't happen (for weeks, so I think it's safe to say it).
2. Mouse cursor moves, but nothing related to mouse event happens (e.g. no switching of focused windows).

This doesn't happen when the network is good, i.e. queries to the filesystem returns reasonably fast. But when there are issues with the network or the remote machine, thie becomes observable and annoying.

Somewhere on the Internet (I forgot the source) says I can ignore some filesystems by adding a configuration `options  = "---exclude-type=nfs4"`, but this doesn't work, and doesn't exist on the wiki. So I believe this option has been deleted.

Therefore, there seems to be no way to avoid quering all mounted filesystems, so supporting async should be very useful, especially when the network latency is big.

------
Although probably doesn't matter, here are the required command outputs:

```
awesome v4.3 (Too long)
 • Compiled against Lua 5.3.5 (running with Lua 5.3)
 • D-Bus support: ✔
 • execinfo support: ✔
 • xcb-randr version: 1.6
 • LGI version: 0.9.2
```

```
Lua 5.3.5  Copyright (C) 1994-2018 Lua.org, PUC-Rio
```

mphe:
Fixes #462

### DIFF
--- a/widget/fs.lua
+++ b/widget/fs.lua
@@ -68,7 +68,7 @@ local function factory(args)
         }
     end
 
-    function fs.update()
+    local function update_synced()
         local pathlen = 10
         fs_now = {}
 
@@ -129,6 +129,10 @@ local function factory(args)
         end
 
         fs.notification_preset.text = tconcat(notifytable)
+    end
+
+    function fs.update()
+        Gio.Async.start(update_synced)()
     end
 
     if showpopup == "on" then


lcpz:
@mphe Thank you for your help.

@renyuneyun Can you test this, please?

mphe:
Forgot to mention: I also had short freezes every now and then. Tested this fix for about two weeks now and had no more freezes so far.

widget/fs.lua
```diff
@@ -131,6 +131,10 @@ local function factory(args)
         fs.notification_preset.text = tconcat(notifytable)
     end
 
+    function fs.update()
+        Gio.Async.start(update_synced)()
```
psychon, line 135: I would suggest to use `Gio.Async.start(gears.protected_call)(update_synced)`.
Try adding `error("shit happens")` somewhere in `update_synced`. Before this commit, AwesomeWM should complain about an error in a callback and continue along. After this commit, Lua will abort the process due to an unprotected error. With my change, `gears.protected_call` we are back at an error message being printed and things continue to chuck along.

mphe: Good to know, I'll fix it.
I don't know how well it works to add commits to an already merged pull request. Should I just open a new one?

lcpz: > Should I just open a new one?

Yes.

psychon: > Shouldn't it be gears.protected_call.call, or did I get something wrong?

It does not really matter since both code paths do the same thing:
https://github.com/awesomeWM/awesome/blob/538586c170cffebca405c005b63f68b125f3d340/lib/gears/protected_call.lua#L47-L56

mphe: > It does not really matter since both code paths do the same thing

@psychon Well, I'm getting an error when using just `gears.protected_call` with `Gio.Async.start`.
It's probably doing some type checking, without considering it being callable.
```lua
local gears = require("gears")
Gio.Async.start(gears.protected_call)(update_synced)
```
```
error while running function!
stack traceback:
	[C]: in function 'coroutine.create'
	/usr/share/lua/5.3/lgi/override/Gio.lua:57: in field 'start'
	/home/marvin/.config/awesome/lain/widget/fs.lua:138: in local 'func'
	/usr/share/awesome/lib/gears/object.lua:148: in function 'gears.object.emit_signal'
	/home/marvin/.config/awesome/lain/helpers.lua:97: in function 'lain.helpers.newtimer'
	/home/marvin/.config/awesome/lain/widget/fs.lua:151: in function 'lain.widget.fs'
	/home/marvin/.config/awesome/widgets/fs.lua:69: in function 'widgets.fs.create'
	/home/marvin/.config/awesome/widgets/base.lua:12: in function 'widgets.fs.allocate'
	(...tail calls...)
	/home/marvin/.config/awesome/rc.lua:243: in main chunk
error: /usr/share/lua/5.3/lgi/override/Gio.lua:57: bad argument #1 to 'create' (function expected, got table)
2020-11-29 14:19:01 E: awesome: Error during a protected call: /home/marvin/.config/awesome/lain/widget/fs.lua:141: attempt to call a table value (upvalue 'callback')
```

psychon: > It's probably doing some type checking, without considering it being callable.

Ah, sorry. I did not know.

PR #473, opened 2020-11-29
Fix async update calls in fs widget

mphe:
Based on the discussion in #471, use `gears.protected_call` in order to get errors logged.
Also found another bug, where `fs.show` is not waiting for `update` to complete.


lcpz:
Thank you.

PR #474, opened 2020-11-29
Fix various luacheck warnings

staticssleever668:
Using _"luacheck . --ignore={111,112,113,631}"_ this reduces warning count from 81 to 9

Ignored warnings:
* 111 - Setting an undefined global variable.
* 112 - Mutating an undefined global variable.
* 113 - Accessing an undefined global variable.
* 631 - Line is too long.

 [Full list of warnings](https://luacheck.readthedocs.io/en/stable/warnings.html)

### DIFF
--- a/helpers.lua
+++ b/helpers.lua
@@ -109,7 +109,7 @@ end
 -- @return cmd PID
 function helpers.async(cmd, callback)
     return spawn.easy_async(cmd,
-    function (stdout, stderr, reason, exit_code)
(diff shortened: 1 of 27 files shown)


lcpz:
Interesting, you have removed warnings in [`dkjson`](http://dkolf.de/src/dkjson-lua.fsl/home) too!

Thank you for your work.

You made me want to remove the remaining ones as well. I will open a new Issue for that.

PR #480, opened 2020-12-18
Remove use of global variables in widget/{alsabar,cpu,mem}.lua

staticssleever668:
This is bad, instead we should pass the data as arguments to `settings` function. I have no motivation to refactor every single widget, but you can see the way I would do it here: 8aabc07d654ca929a67442eb79e624b7a584ef80.

*Old text:*
#### This introduces a new way of changing 'settings()' function of `cpu` and `mem` widgets:
Instead of setting mem's `settings()` function like this: 
```
local cpu = lain.widget.cpu({
       settings = function()
               widget:set_markup(string.format("%s%%", cpu_now.usage))
       end,
})
```
Following can be used:
```
local widget_ram = lain.widget.mem()

widget_ram.settings = function()
	widget_ram.widget:set_markup(string.format("%s MB (%s%%)", widget_ram.used, widget_ram.perc))
end
```
*Edit: this can be made even better by passing these members as arguments to settings() function.*
In fact, this is required now, as the global variables in the first example are no longer defined
`timeout` is now also a member but it can't be changed like this because it's used outside of `update` function.
TODO: things like `timeout` should not be members. References to members available in the `settings` function should be it's arguments.

#### This would also require following changes to wiki:
To **alsabar:**
In the table after `settings can use the following variables:`:
Replace `volume_now.level` with `volume._current_level` 
Replace `volume_now.status` with `volume._playback`

To **mem**:
Replace `mem_now` occurrences with `mymem`

To **cpu**:
Replace `cpu_now.usage` with `cpu.core[0].usage`
Replace `cpu_now` occurrences with `cpu.core`

To both **mem** and **cpu**:
Indicate that `settings` and `timeout` are now available as instance variables

To the **index page of "widgets" section**:
Change the example to this:
```
local cpu = lain.widget.cpu()
cpu.settings = function()
    widget:set_markup("CPU " .. cpu.core[0].usage)
end
```

#### Additional note
Maybe I missed something and something else has to be changed as well.
The changes are breaking, but global variables should not be used, and this can be the first step to get rid of them.
There are also comments under commits, maybe they are worth checking out. Commit messages could be too long, tell me if it matters.

### DIFF
--- a/widget/alsabar.lua
+++ b/widget/alsabar.lua
@@ -95,11 +95,6 @@ local function factory(args)
                     alsabar.bar.color = alsabar.colors.unmute
                 end
 
-                volume_now = {
-                    level  = alsabar._current_level,
-                    status = alsabar._playback
-                }
-
                 settings()
 
                 if type(callback) == "function" then callback() end
--- a/widget/cpu.lua
+++ b/widget/cpu.lua
@@ -18,8 +18,9 @@ local function factory(args)
     args           = args or {}
 
     local cpu      = { core = {}, widget = args.widget or wibox.widget.textbox() }
-    local timeout  = args.timeout or 2
-    local settings = args.settings or function() end
+
+    cpu.timeout  = args.timeout or 2
+    cpu.settings = args.settings or function() end
 
     function cpu.update()
         -- Read the amount of time the CPUs have spent performing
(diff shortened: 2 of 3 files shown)


lcpz:
May I know why you close? If it is for me, I am still busy and this PR is quite big, but I would have given it a look as soon as I could.

staticssleever668:
Hi, I closed this because I think that this implementation is bad. Instead of getting data from widget members (and because of that having to set `settings` separately from widget constructor), I think we should pass the data as arguments to `settings` function. I only use one widget where global variables are used like that, so I only had motivation to fix that one widget. Here is how I did it for reference: 8aabc07d654ca929a67442eb79e624b7a584ef80.

PR #518, opened 2021-10-13
collect max fs usage in percent

lexxxel:
### Suggestion
Add a variable that knows the maximum filled partition in percent.

### Reason
fs_now["/"].used showed strange numbers on my system. I also found it useful to know the max used partition value so I added a variable for that. 
  
I did this for my system and liked the change, I don't mind if this PR does not fulfil your criterias and you simply close it.

### DIFF
--- a/widget/fs.lua
+++ b/widget/fs.lua
@@ -74,6 +74,7 @@ local function factory(args)
     local function update_synced()
         local pathlen = 10
         fs_now = {}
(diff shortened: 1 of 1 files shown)


lcpz:
The usefulness of metrics like `max` and `min` is completely arbitrary. Also, such practice is not KISS.

I'd suggest you move this logic in your `rc.lua`.

PR #520, opened 2021-11-21
Uptime widget

nenikitov:
Implemented a widget that gets the info about system's uptime.

### DIFF
--- a/widget/contrib/uptime.lua
+++ b/widget/contrib/uptime.lua
@@ -1,0 +1,44 @@
+--[[
+
+     Licensed under GNU General Public License v2
+      * (c) 2021, nenikitov <https://github.com/nenikitov>
+
+--]]
+
+local helpers              = require("lain.helpers")
+local wibox                = require("wibox")
+local match, lines, floor  = string.match, io.lines, math.floor
+
+local function factory(args)
+    args           = args or {}
+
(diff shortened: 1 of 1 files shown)


lcpz:
Thank you for your interest in my repository, and for taking the time to write this widget.

Unfortunately, I cannot accept it because what you want can be easily achieved with a [`watcher`](https://awesomewm.org/recipes/watch/) like this:

```lua
local floor = math.floor
local uptime = awful.widget.watch(
    'bash -c "cat /proc/uptime | cut -d\' \' -f1"',
    1, -- refresh rate (in seconds)
    function(widget, stdout)
        local seconds = floor(stdout)
        local minutes = floor(seconds / 60) % 60
        local hours   = floor(seconds / 3600) % 24
        local days    = floor(seconds / 86400)

        seconds = seconds % 60

        -- customize here
        widget:set_text(string.format(
            "Uptime: %02d:%02d:%02d:%02d", days, hours, minutes, seconds)
        )
    end
)
```

The reason for prioritising watchers is that `awful.widget.watch` is supported officially, while my repository is not.

Feel free to add this code to the [Recipes section](https://github.com/awesomeWM/awesome-www#contributing-to-recipes-section) of the AwesomeWM website.

PR #523, opened 2022-03-12
Add format option for net and temp widget

wahlencraft:
This lets the user specify format for the string in the net and temp widget. For example to change the number of decimals.

I did not understand how to add to the wiki (do I fork the wiki somehow?). I was thinking adding this line (to both widgets) would be enough:
```
`format` | String format for output | string | "%.1f"
```


lcpz:
I added that line to the wiki pages of `net` and `temp`.

PR #525, opened 2022-03-22
Fixed require statements

Trial97:
The `require` statements should be dynamic in order to place this project in another folder other than the root config folder

### DIFF
--- a/widget/bat.lua
+++ b/widget/bat.lua
@@ -6,7 +6,12 @@
 
 --]]
 
-local helpers  = require(tostring(...):match(".*lain") .. ".helpers")
+local requirePrefix = tostring(...):match(".*lain") or ""
+if requirePrefix then
+	requirePrefix = requirePrefix .. "."
+end
+
+local helpers  = require(requirePrefix .. "helpers")
 local fs       = require("gears.filesystem")
 local naughty  = require("naughty")
 local wibox    = require("wibox")
--- a/widget/bat.lua
+++ b/widget/bat.lua
@@ -6,7 +6,7 @@
 
 --]]
 
-local helpers  = require("lain.helpers")
+local helpers  = require(tostring(...):match(".*lain") .. ".helpers")
 local fs       = require("gears.filesystem")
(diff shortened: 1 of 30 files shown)


widget/bat.lua
```diff
@@ -22,7 +22,7 @@ local function factory(args)
     local pspath = args.pspath or "/sys/class/power_supply/"
 
     if not fs.is_dir(pspath) then
-        naughty.notify { text = "lain.widget.bat: invalid power supply path", timeout = 0 }
+        naughty.notify { text = tostring(...):match(".*lain") .. ".widget.bat: invalid power supply path", timeout = 0 }
```
lcpz, line 25: `naughty.notify` is not vararg, so this raises the following error:

```lua
cannot use '...' outside a vararg function near '...'
```

Since we cannot touch `naughty.notify`, please revert this change.

lcpz:
Thank you for your work. Please see my review.

Before merging, can please you ensure that you tested every widget, layout and utility?

Maybe you can help me with #527 as well.

Trial97:
Thanks for the review. I reverted the change.
Regarding testing everything it will take me some time as I just started learning lua and awesomewm.
As regards #527 can you elaborate?
It is enough for the files to only have the `require` statement in them?
e.g.lain-test-quake.lua
```
local lain      = require('extra.lain')
return lain.util.quake({
    app = 'kitty',
    name = "QuakeDD",
    argname = "--name %s",
    followtag = true,
    settings = function(c)  ...    end
    extra = '',
    border = 1,
    visible = false,
    overlap = false,
    screen = awful.screen.focused(),
    height = 0.25,
    width = 1,
    vert = "top",
    horiz = "left"
})
```

Trial97:
Ok, a small update to this:
- added a first set of tests(not all files but maybe)
- fixed requires to work with both  tests and also in an awesome config

Let me know if there are some changes that need to be done regarding tests(if not I will continue adding them in the same manner if I have time)

[docs](https://lunarmodules.github.io/busted/) for the future :D

Issue #527, opened 2022-03-24
We need a test suite


lcpz:
The idea is to have a `lain-test.lua`, which requires `lain-test-{util, widget, layout}.lua`, each of which requires all elements of the respective submodule with all possible options set. Tests could be performed with Xephyr as follows:

```shell
Xephyr :1 -screen 1280x800 2> stdout.txt & DISPLAY=:1 awesome -c lain-test.lua
```

This structure and workflow will serve as a unit, integration and system test suite.

They will also be used to do regression testing for each new commit (e.g., using Travis CI).

thomascft:
Ok, so... I wrote a script with Github Actions, but I'm not sure how to implement the widget tests. Right now i just `require("lain.widget")`. Does that enumerate through the widgets below it or not. I could just make an array with all the widget's names but It'd be important to update that list whenever a new one is added.

Aire-One:
Hello, 

That's nice to see some effort in this direction. The more the Awesome WM ecosystem (awesome itself and third-party modules / dependencies) has automatic test, the more it can be considered as mature and stable.

Regarding our tests at Awesome, we have a similar test suite for our integration tests than the one you propose here : https://github.com/awesomeWM/awesome/blob/master/tests/run.sh. It starts Xephyr and run step-by-step tests withing a graphical instance of Awesome. I think, the biggest downside of it, is that it's a homegrown tool, so not exactly what you expect from a conventional/standardized tool, and you have to learn it from scratch. It's not hard, but I'm not sure how easy it would be to reuse a similar script on another context, like for the lain module.

Unit testing is done with busted (https://olivinelabs.com/busted/). Because unit tests don't load a full working environment with Awesome, but only the part that is tested, we need to mock dependent modules in our tests, and to have this preload script (https://github.com/awesomeWM/awesome/blob/master/spec/preload.lua). It's a common annoyance of unit testing that you can't avoid. I'm also not sure how bad it would be for third-party modules.

(just my 2cts about CI testing Awesome WM)

PR #531, opened 2022-06-08
Made paths relative

thomascft:


thomascft:
It appears there is already another open request accomplishing similar things. If that's still being considered I can close this, if not I have tested everything, and as far as I can tell it works.

lcpz:
Yes, already considered in #525.

Did you do your tests manually? Could you help with #527 so to make them automated?

thomascft:
I don't have to a of experience, but I'd be willing to help.

thomascft:
Well, I don't know git well enough clearly. I setup the testing, you can see the output [here](https://github.com/thomascft/lain/runs/6803123961?check_suite_focus=true). I had a bit of cleanup, so that's what the force pushes were, oh... I'm gonna close the pull request and do it with in reference of #527

PR #533, opened 2022-06-08
Automated testing with Luacheck 

thomascft:
This is my fix for #527.
Well, I don't know git well enough clearly. I setup the testing, you can see the output [here](https://github.com/thomascft/lain/runs/6803123961?check_suite_focus=true). I had a bit of cleanup, so that's what the force pushes were

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -1,0 +1,12 @@
+name: Luacheck
+on: [push]
+jobs: 
+  Luacheck_Tests:
+    runs-on: ubuntu-latest
+    steps:
+      - run: echo "Running tests triggered by a ${{ github.event_name }} event."
+      - run: echo "Testing ${{ github.ref }} from ${{ github.repository }} on ${{ runner.os }}"
+      - name: Check out code
+        uses: actions/checkout@v3
+      - name: Run tests
+        uses: lunarmodules/luacheck@v0 
--- a/.luacheckrc
+++ b/.luacheckrc
@@ -1,0 +1,11 @@
+codes = true
+allow_defined = true
+max_line_length = false
+
+globals = {
+    "client", "screen", "mouse"
+}
+
+ignore = {
+    "131",
+}


thomascft:
I should probably mention that the test rules are defined in `.luacheckrc` with the documentation over [here](https://luacheck.readthedocs.io/).

lcpz:
Thank you for this. However, we still need unit tests.

It's not so difficult now, we just need to create a test script with [luaunit](https://github.com/bluebird75/luaunit) and run it in the Action you created, just after the linter call.

Feel free to continue to help me, I think it's a nice exercise for you.

Otherwise, I will work on it as soon as I can.

.github/workflows/main.yml
```diff
@@ -0,0 +1,12 @@
+name: Luacheck
+on: [push]
+jobs: 
+  Luacheck_Tests:
```
lcpz, line 4: For the next PR, you can add another job, say `unit_tests`, with a step in which you run the testing script.

thomascft: I'll try it out, should be able to get it in by Wednesday.

thomascft: Are you set on luaunit? [Busted](http://olivinelabs.com/busted/) seems to have much better documentation, and the AwesomeWM team said it's what they use in #527. It would be nice to see how they've structured unit testing.

lcpz: Busted is fine. Take inspiration from AwesomeWM.

thomascft: What exactly do you need tested? I haven't done unit testing before and I'm not sure it's being done correctly.

lcpz: We need to launch an Awesome instance in which we load all the widgets (with all possible configurations) to see if they work. This test suite would be called for each commit to verify if everything is OK. I know this is a lot of work, but you can start by launching Awesome as mentioned [here](https://github.com/lcpz/lain/issues/527#issuecomment-1152897552=). We can incrementally add the widget tests.

PR #540, opened 2022-10-18
fix cal day alignment for short day names

hp-pepster:
If the day names do not consist of three characters, the alignment is off (see https://github.com/lcpz/lain/issues/522)


widget/cal.lua
```diff
@@ -53,6 +53,15 @@ local function factory(args)
         return (x + y) % 7
     end
 
+	function cal.get_dayname(day_num)
```
lcpz, line 56: Indentation is not correct.

Czarnodziej:
Looks good to me (Lua 5.4.4)

PR #549, opened 2023-01-13
introduce helpers.uri (curl replacement) & migrate widget/weather to it

razamatan:
- new dependency: needs libsoup gi installation (smaller than gvfs)
- [breaking] weather: `(current/|forecast)_call` changed to `(current|forecast)_uri`
- [breaking] weather: use scoped (non-global) `now.current` and `now.forecast` to avoid `weather_now` clobbering
- weather: display something useful by default for the top-level widget
- weather: render the hour of the forecast by default since the default forecast is 5 days by 3 hour increments
- weather: add degree symbol to temperatures

### DIFF
--- a/helpers.lua
+++ b/helpers.lua
@@ -7,6 +7,7 @@
 
 local spawn  = require("awful.spawn")
 local timer  = require("gears.timer")
+local gdebug = require("gears.debug")
 local debug  = require("debug")
 local io     = { lines = io.lines,
                  open  = io.open }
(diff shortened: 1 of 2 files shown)


lcpz:
Pending discussion in #349.

> [breaking] weather: use `current` and `forecast` to avoid `weather_now` clobbering

This makes me realise that we should avoid using global variables. #550

razamatan:
re: globals.  yeah, i'll break out the breaking global var change into a different pr.

razamatan:
see https://github.com/lcpz/lain/pull/349#issuecomment-1396447885. i'm going to go ahead and break out the global change to a later pr to unblock this one.

razamatan:
updated.  went ahead and did the global state localization for the weather widget in this pr since this is a breaking change anyway.  i'll have different PRs for some of the other widgets i mess with.

razamatan:
ping

PR #564, opened 2025-09-27
Fix issue for GIO - Fs widget not working

madhur:


sleeptightAnsiC:
Thanks! For reference, I mentioned the issue that this PR fixes here: https://github.com/lcpz/lain/issues/563#issuecomment-3339833675

I just wonder if the fix should _**support both "versions" of glib**_ (?) I'm not very familiar with glib so maybe I'm talking nonsense right now... I only raised this concern because AwesomeWM upstream mentioned something similar here https://github.com/awesomeWM/awesome/pull/4022#discussion_r2372619838 and there is alternative upstream PR that tries to support both version https://github.com/awesomeWM/awesome/pull/4027

lcpz:
Thanks for your prompt fix.

Would it be possible to add support for both versions?

madhur:
Updated

lcpz (approved):

